PR #29, opened 2018-07-11
Fix obj loader

rdeits:
cc @gizatt

This was accidentally breaking the first line of every .obj file. But since most obj files start with a comment or material description, I never noticed.


gizatt:
Wow, nice find!

PR #30, opened 2018-07-23
Shadows

gizatt:
A little cludgy, as it enables shadows (casting + receiving) for all bodies. Future work would be to make sure that's toggleable on a per-object level over the message passing. (Doesn't seem to work right now, but I've only given it a cursory attempt.)

But much prettier!

image


gizatt:
Updated PR to disable shadows by default. Currently disabling them at the light level (as a toggleable option on each spotlight), instead of at the global level that we discussed. Functionally equivalent, I hope, but required less tweaking from the original PR... happy to switch to a global shadow switch if you think it's better, but shadow control on a per-light level seems nice anyway.

rdeits:
Sounds great. Thanks!

PR #41, opened 2018-11-10
add `FontLoader` parsing, and add font JSON to local assets

shensquared:
still WIP
FYI, the `fonts` folder are directly copied from the three.js git repo, first tried remote loading the font js file but git denies access, hence add it local. Not entirely sure about the license but seems OK?


rdeits:
Cool! I think before we go further down this road, we should make sure that TextGeometry is actually the right thing. Do you want your text to be an extruded 3D object (like the demo here: https://threejs.org/docs/#api/en/geometries/TextGeometry ) or just 2D text on a flat surface? 

If 2D text is enough, then I bet we could come up with a simpler approach by just drawing text onto a Canvas and then setting that as a texture (option 2 from https://threejs.org/docs/#manual/en/introduction/Creating-text ). That would avoid needing to ship our own fonts, since we could just do something like: 

```js
let canvas = document.createElement('canvas');
canvas.width = 400;
canvas.height = 200;
let ctx = canvas.getContext('2d');
ctx.font = '48px sans-serif';
ctx.fillText('Hello world', 10, 10);
```

and then set that canvas as the texture of a flat surface. 

The advantages of the 2D approach are that we don't have to worry about shipping fonts (the browser already knows how to render text, after all) and it would be easy to do things like put text onto some object other than just a flat surface. The advantages of the 3D approach are that it looks awesome and makes the text a real 3D part of the scene. 

Are you relying on the 3D-ness of the text in your proposed implementation?

shensquared:
:woman_facepalming: you know what, the 2d option somehow completely escaped me… 

Indeed, all I need is to have descriptions aligned (time-wise) with the animations so to avoid post manual editing. 2D is suffice, and come to think of it, perhaps even preferable for the task (3D text might be distracting). Will definitely go with this, thanks Robin. 

I’ll close all the PR and issues then?

rdeits:
No problem! Let's at least keep the issues open, and I'm happy to keep helping with option 2. You're definitely not the first person to want text rendering, and it would be great to have.

shensquared:
close since we are attempting a new approach in #43

PR #42, opened 2018-11-11
Switch back to directional light by default

rdeits:
@gizatt I've been having a variety of issues getting things to look right for a variety of scenes with the SpotLight as the primary light source, especially for scenes with moving objects that can leave whatever area the spotlight is focused on. So I'd like to switch back to using a DirectionalLight by default. To make it easy to switch back and forth, I figured I could just load both lights, but set the spotlight's visibility to `false` by default. 

I want to make sure that it's still easy for you to render things the way you want after this change (with spotlight and shadows). It should just be a matter of doing: 

```py
vis["/Lights/DirectionalLight"].set_property("visible", False)
vis["/Lights/SpotLight"].set_property("visible", True)
vis["/Lights/SpotLight/<object>"].set_property("castShadow", True)
```

perhaps with some additional adjustments to intensity. Ultimately it's going to be impossible to choose one lighting setup for every use case, but I'd like to have the defaults be unobjectionable and make customization as easy as possible.


rdeits:
As an example, here's a field of cats on a flat surface before this change: 

meshcat-2

and after: 

meshcat

gizatt:
Yeah, that's totally fine with me! Sorry the switch caused trouble in the first place.

PR #48, opened 2019-02-14
Add support for DAE and STL meshes

wxmerkt:
Hi Robin,
Thank you for creating this awesome visualiser! We've been thinking of using it for a while and are currently integrating it with Pinocchio and Exotica.

This PR adds support for loading DAE and STL meshes by using the three.js loaders for these types. 

Note, I've copied them into `src/` the same way you added the OBJLoader2 but I am not sure that is even necessary - and there are likely better ways to do it (in webpack?) - very open for suggestions here :-).

I'll open the corresponding `meshcat-python` PR shortly.

Thank you,
Wolfgang

### DIFF
--- a/src/ColladaLoader.js
+++ b/src/ColladaLoader.js
@@ -1,0 +1,3918 @@
+/**
+ * @author mrdoob / http://mrdoob.com/
+ * @author Mugen87 / https://github.com/Mugen87
+ */
+
+THREE.ColladaLoader = function ( manager ) {
+
+	this.manager = ( manager !== undefined ) ? manager : THREE.DefaultLoadingManager;
+
+};
+
+THREE.ColladaLoader.prototype = {
+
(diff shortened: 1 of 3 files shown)


rdeits:
This looks great, thank you! Copying the loaders into the source tree is fine: it would also be possible to wrap them up in node modules, but I don't think it's necessary or worth the effort. 

One thing to note: the generated file `dist/main.min.js` is intentionally checked into this repo, so I'm going to merge this PR and then make one more commit on master to update that generated file.

wxmerkt:
Thank you, Robin! I'll wait for your commit with the `dist/main.min.js` to update the `meshcat-python` PR in the morning - and will add two meshes as examples (I tested with locally available ones).

rdeits:
Ok, done: https://github.com/rdeits/meshcat/commit/fc7e50d232a866129c8dd82b4217335158f3dcf6

PR #51, opened 2019-05-19
Slightly hack-y workaround for .DAE loading

rdeits:
A step towards fixing https://github.com/JuliaRobotics/MeshCatMechanisms.jl/pull/32

With this PR, I think .DAE MeshFile objects should work pretty well. The only issue is that it's possible for the user to try to specify a material in Julia for a .DAE MeshFile, but that specification will have no effect on what's rendered. I think the right fix is to not let users specify materials for MeshFiles anyway (as in https://github.com/JuliaRobotics/MeshCatMechanisms.jl/pull/32#issuecomment-493781957 ), so this is OK on the meshcat javascript side.


rdeits:
@tkoolen Here's the forearm .DAE file with this PR:

meshcat

tkoolen:
OK, this seems great to me. Good to merge?

rdeits:
I'm actually working on a better solution right now. I'm going to do the slightly harder but more correct thing of treating a mesh file as an entire Object, not just a geometry. That requires some extra work to handle .obj files (because they annoyingly store their materials in separate files), but it will make this code more correct and should mean that, for example, Valkyrie will now show up with textures.

rdeits:
Closing in favor of #54

PR #54, opened 2019-05-29
Implement mesh loading for objects and geometries

rdeits:
Supersedes #51 

This PR implements loading mesh files both as Geometry (using the `type: _meshfile_geometry` attribute) and as Object (using the `type: _meshfile_object`) attribute. The idea is that users can choose whether to treat the file as a plain geometry source or as a complete object with its own materials and textures. It also includes support for sending auxiliary files (like textures) as data URIs so that the browser can access what would otherwise have to be local files. 

This will need a little work to support on the Julia/python sides. It's backwards compatible, but we need to modify the clients to take advantage of the ability to send textures and materials.


tkoolen:
Do you need anything from me for this?

rdeits:
No, sorry--I was just stalled on one lingering issue (ensuring the view actually re-draws after asynchronously loading an image texture). I think we're good now.

PR #55, opened 2019-07-01
update all js dependencies

rdeits:
Fixes some security vulnerabilities in a couple of js packages.

PR #57, opened 2019-07-13
Handle multiple geometries in a .dae meshfile_geometry

tkoolen:
On top of #55.

This seems to be pretty slow but it almost works. I'm now getting

image

with https://github.com/tkoolen/universal-robot-dae-jl. Note the missing mesh at the base.


tkoolen:
I think a likely cause for the missing base link geometry is that I'm skipping `child.type == "Group"`. Also I think we'd ideally want to not even use `mergeBufferGeometries` since it appears to be very slow, instead adding the individual geometries to the scene graph. But they each need a unique uuid; they can't just share `geom.uuid`.

src/index.js
```diff
@@ -61,9 +62,21 @@ function handle_special_geometry(geom) {
         } else if (geom.format == "dae") {
             let loader = new THREE.ColladaLoader();
             let obj = loader.parse(geom.data);
-            let result = obj.scene.children[0].geometry;
-            result.uuid = geom.uuid;
-            return result;
+            let buffer_geometries = []
+            for (let child of obj.scene.children) {
```
VladimirIvan: Some collada files have hierarchy deeper than 1 layer. The base link may be one of them.

src/index.js
```diff
@@ -61,9 +62,21 @@ function handle_special_geometry(geom) {
...
+                if (child.type == "Mesh") {
+                    let geometry = child.geometry;
+                    if (geometry.type == "BufferGeometry") {
+                        buffer_geometries.push(geometry);
```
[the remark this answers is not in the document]
tkoolen: Ah, you're right. At first I was looking at [`mergeMesh`](https://threejs.org/docs/index.html#api/en/core/Geometry.mergeMesh), which automatically applies the matrix.

src/index.js
VladimirIvan: I think we may have implemented the same feature in parallel. I was just finishing it up when you created the PR :)

tkoolen: If yours is in a better state, please do open it!

VladimirIvan: I added texture and material loading as well.
See #58.

tkoolen:
Closing in favor of #58.

PR #61, opened 2019-09-17
Add more shadow controls, and change DirectionalLight to PointLight

gizatt:
The main fix here is fixing some shadow banding issues I was seeing by increasing default shadow map size a lot (hopefully not too much? it's only 4mb, but I have a beefy computer), and most importantly, adding a little shadow bias.

I'm not super attached to the DirectionalLight->PointLight change -- it works well for *my* scene scales (handfuls of meters), but I dunno what other folks use. The `distance` control can be used to get light up to pretty large distances (with 0 distance supposed to be infinite lighting, so functionally the same as DirectionalLight? But I'm not sure I believe it after testing it).

Before, with and without shadows:
image
image

After, with and without shadows:
image
image

PointLights are particularly nice because they support falloff that adds some depth cues to otherwise flat scenes. The "good" settings are very scene-dependent, but as an example for this scene, these are the best parameters I could find tweaking only the parameters made adjustable in this PR:
image

---
This change is [Reviewable](https://reviewable.io/reviews/rdeits/meshcat/61)


rdeits:
This is really cool! I think it's definitely an improvement. Let me just run a few tests locally to make sure the things I'm doing work with the parameters you've picked.

ferrolho:
Bump! This looks great! 🧐

ferrolho:
Is there something holding this back, @rdeits ?

rdeits:
@gizatt mentioned in person that he had some more changes he wanted to make--is that still the case?

gizatt:
Uhh... whatever they were, I've forgotten about them, so I'm happy with this going in as-is! (It's what I'm using on my own machine right now and I'm satisfied with it.)

ferrolho:
:bump: @rdeits

rdeits:
This is awesome. I've merge master into it over in #66 so I'll merge that PR. Thanks for making this happen!

ferrolho:
Thank you both! This is really good.

PR #69, opened 2020-04-16
Import BufferGeometryUtils from three.js examples

gizatt:
I *think* this is necessary -- I'm seeing an error

```
Uncaught TypeError: Cannot read property 'mergeBufferGeometries' of undefined
at merge_geometries (main.min.js:35)
```

otherwise.

---
This change is [Reviewable](https://reviewable.io/reviews/rdeits/meshcat/69)


rdeits:
Oh, I see--I think you're right, and this makes sense (as much as anything in javascript packaging makes sense...). Thanks for the fix!

EricCousineau-TRI:
I'm not sure if this was fixed for all browser? :(
https://github.com/RobotLocomotion/drake/pull/13050#pullrequestreview-400430779

Not sure what the proper fix is here. I don't think I had any stale deps (I rebuilt in Bazel, etc.).

EricCousineau-TRI:
Nevermind. It seems like it's a Firefox caching thing :(

PR #71, opened 2020-07-22
Support SSL in the viewer.

RussTedrake:
It never makes sense to use ws: when connecting with the https: protocol.  The follow-up PR will add SSL to the zmqserver in meshcat-python.


RussTedrake:
FTR, you can preview the proposed changes to the zmqserver [here](https://github.com/rdeits/meshcat-python/compare/master...RussTedrake:ssl?expand=1).

PR #75, opened 2020-07-26
Fix orthographic camera

### LINKED ISSUE #3: Support orthographic camera
It would be nice to make orthographic projections easy to create and still controllable with the mouse. 

The current workaround is: (in the Javascript console):

```js
camera = new THREE.OrthographicCamera(-2, 2, 2, -2, -2, 2);
```

and then to use `camera.rotateY(Math.PI / 2)` to rotate the camera. 

cc @pangtao22

RussTedrake:
And demonstrate that it can be constructed via the handle_command interface.
This example implements essentially a "planar visualizer" (in the language of drake).

Resolves #3

---
This change is [Reviewable](https://reviewable.io/reviews/rdeits/meshcat/75)


RussTedrake:
___
*[test/orthographic_camera.html, line 64 at r1](https://reviewable.io/reviews/rdeits/meshcat/75#-MDBf2BW-skA5kzsKmkE:-MDBf2BW-skA5kzsKmkF:bq1bv4b) ([raw file](https://github.com/rdeits/meshcat/blob/a1662af6ad4a4feca4531ebfe09fb23b95503ff2/test/orthographic_camera.html#L64)):*
> ```HTML
>                 path: "/Cameras/default/rotated",
>                 object: {
>                     object: {
> ```

btw -- i don't actually really understand why I needed the nested objects here.  is that what you would have expected?

rdeits:
> btw -- i don't actually really understand why I needed the nested objects here. is that what you would have expected?

Yeah, it's correct. This is just a slight weirdness of the threejs API, combined with the fact that I use "object" to mean "thing you want to create" and threejs *also* uses it as a field of the object itself. You can see the same nested `object` in the `set_object` API documentation in the Readme.

rdeits:
Looks great, thanks!

PR #79, opened 2021-01-20
[Draft] Adds handling of "take_screenshot" commands

ferrolho:
This goes in hand with [rdeits/MeshCat.jl#195](https://github.com/rdeits/MeshCat.jl/pull/195).


src/index.js
```diff
@@ -985,6 +986,9 @@ class Viewer {
             this.set_animation(cmd.animations, cmd.options);
         } else if (cmd.type == "set_control") {
             this.set_control(cmd.name, cmd.callback, cmd.value, cmd.min, cmd.max, cmd.step);
+        } else if (cmd.type == "take_screenshot") {
```
rdeits, line 989: Could you rename this to `save_image` as well, to reflect the fact that the image is saved to disk? Someday I'd also like to enable transmitting the image back to the Julia client, so I'd like to make sure that the behavior implemented here has a clearly identifiable name.

rdeits (commented):
Thanks! Just one minor request.

ferrolho:
Done :slightly_smiling_face:

rdeits:
Thanks!

>Actually, should I have committed dist/main.min.js? 

Yes--it's annoying to have to commit it manually, but it makes this package easier to use downstream.

PR #85, opened 2021-07-30
Feature: command to query image from client

ManifoldFR:
This PR adds a command `capture_image` to query the viewer for a PNG image of the rendered view. This returns a JSON object
```json
{
  "type": "img",
  "data": "data:image/png;base64,XXXXX"
}
```


rdeits (approved):
Looks great, thanks!

src/index.js
```diff
@@ -985,6 +985,12 @@ class Viewer {
             this.set_animation(cmd.animations, cmd.options);
         } else if (cmd.type == "set_control") {
             this.set_control(cmd.name, cmd.callback, cmd.value, cmd.min, cmd.max, cmd.step);
+        } else if (cmd.type == "capture_image") {
+            let imgdata = this.capture_image();
+            this.connection.send(JSON.stringify({
```
RussTedrake, line 990: Could we use msgpack here instead?  It seems we shouldn't be using two different, almost identical representations on the same websocket?

PR #88, opened 2021-08-18
minor tweaks to support headless testing

RussTedrake:
1) Allows one to optionally pass a renderer to the Viewer constructor
2) Count the number of messages received
3) Use window.CCapture instead of CCapture, as recommended in
https://stackoverflow.com/questions/48864251/how-to-run-ccapture-by-server-side-node-js
(and it still works in the normal browser)


RussTedrake:
Related to https://github.com/RobotLocomotion/drake/issues/13038.

src/index.js
```diff
@@ -689,12 +689,16 @@ function gradient_texture(top_color, bottom_color) {
...
-        this.renderer.shadowMap.enabled = true;
-        this.renderer.shadowMap.type = THREE.PCFSoftShadowMap;
-        this.dom_element.appendChild(this.renderer.domElement);
+        if (renderer === undefined) {
+            this.renderer = new THREE.WebGLRenderer({antialias: true, alpha: true});
+            this.renderer.shadowMap.enabled = true;
+            this.renderer.shadowMap.type = THREE.PCFSoftShadowMap;
+            this.dom_element.appendChild(this.renderer.domElement);
+        } else {
+            this.renderer = renderer;
```
gizatt, line 700: If it's possible, asserting `renderer` is the right type here might prevent really confusing deferred errors later if someone shoves in a bad renderer type (e.g. accidentally invokes the constructor with a weird extra argument)?

gizatt (approved):
Took a look and looks good for merge with a one nit, and with one caveat: the ccapture change is pretty mysterious to me, but I trust it fixed a problem for you, and it looks innocuous as long as it still runs and passes tests.

src/index.js
rdeits: That's a good point, although I don't actually know what the right assertion would be, since javascript typing is kind of the wild west. I'm ok with leaving this open-ended for now.

rdeits (approved):
:+

rdeits:
By the way, I stalled on updating the `meshcat` submodule in `meshcat-python` to `master` over in https://github.com/rdeits/meshcat-python/pull/94 because it seems that a recent update to Three.js itself caused some flickering when running an animation that involved simultaneously changing the camera zoom and some other property of the scene. I think I'll probably just declare that to be a known issue and go ahead with the upgrade, since I don't want to block new features forever (and since animating camera zoom is not exactly a critical feature anyway).

PR #94, opened 2021-08-28
Add set_control_value and delete_control

RussTedrake:
Before, we can create controls with the command set_control, but had no way to delete them.


rdeits (approved):
Looks good--thanks for adding the test!

PR #98, opened 2021-08-30
Add option to set_control_value without invoking the callback

RussTedrake:
In a complicated workflow, like synchronizing sliders across multiple meshcat browsers, the ability to set the value without triggering the callback becomes quite useful.

We simply shortcut the callback call in https://github.com/dataarts/dat.gui/blob/master/src/dat/controllers/Controller.js#L89


src/index.js
```diff
@@ -995,11 +995,15 @@ class Viewer {
         }
     }
 
-    set_control_value(name, value) {
+    set_control_value(name, value, invoke_callback=true) {
```
rdeits, line 998: Believe it or not, default values are a relatively recent (2015-ish) feature in javascript. I'm glad we don't have to do `if (invoke_callback === undefined)` anymore...

rdeits (approved):

PR #100, opened 2021-09-05
fix dat.GUI button elements layout

RussTedrake:
without this fix, buttons added to the GUI with a long name were artificially truncated to the same length that a slider name gets truncated (to make room for the slider and text input).  The button should not have that restriction.

After this fix, the complete long name of the button is displayed properly.


test/controls.html
```diff
@@ -35,6 +35,10 @@
             // The slider is updated, but the javascript console confirms that 
             // the callback was not called.
             viewer.set_control_value("TestSlider", .4, false);
+
+            viewer.set_control("Test button with a really long name", () => {
```
rdeits, line 39: Can you change this file to include `main.min.js`? It currently uses `main.js` which isn't tracked in git

RussTedrake: Done.  (I had copied it from empty_scene.html)

RussTedrake:
FWIW, I've also gone ahead and PR'd the upstream fix.  https://github.com/dataarts/dat.gui/pull/306 .  If that goes through, then we can bump the dependency and get rid of the one-line bandaid.  But I would like to land this for now.

test/controls.html
rdeits: thanks!

PR #102, opened 2021-09-18
Set correct device pixel ratio for rendering

ManifoldFR:
On HiDPI screens, we need to use the window object's `devicePixelRatio` to set the renderer's pixel density to the screen's detected pixel density. This avoids getting blurry output being rendered to the viewpoint on e.g. retina displays.


rdeits:
Thanks! I don't have a retina display to test this on, but it seems to still work fine on non-retina devices, and I'll take your word that this helps out for HiDPI.

PR #111, opened 2022-01-11
Set resolution of captured images & change camera target

ManifoldFR:
Hi,

This PR adds two features:

- as a follow-up to #85, be able to ask for a specific render resolution for the image (this is implemented by resizing the pane before & after rendering and getting the image DataURL) which is nice for people with smaller screens; by default this is set to 1920x1080.
- expose a function to set the camera target over the programmatic JS/WebSocket API, which is a bit more convenient than getting the right 3D rotation to point to where you want to look at (this is implemented by changing the `.target` attribute on the `Viewer.controls` object)

I updated the README accordingly. A follow-up PR on rdeits/meshcat-python will be coming soon


src/index.js
```diff
@@ -1040,14 +1049,20 @@ class Viewer {
...
         } else if (cmd.type == "set_control_value") {
             this.set_control_value(cmd.name, cmd.value, cmd.invoke_callback);
         } else if (cmd.type == "delete_control") {
             this.delete_control(cmd.name);
         } else if (cmd.type == "capture_image") {
-            let imgdata = this.capture_image();
+            let w = cmd.xres || 1920;
+            let h = cmd.yres || 1080;
+            w = w / this.renderer.getPixelRatio();
+            h = h / this.renderer.getPixelRatio();
```
ManifoldFR, line 1064: this is needed to account for the automatic upscale due to device pixel ratio (needed for iPads, Macbooks, other HiDPI screens)

src/index.js
```diff
@@ -1040,14 +1049,20 @@ class Viewer {
...
         } else if (cmd.type == "set_control") {
             this.set_control(cmd.name, cmd.callback, cmd.value, cmd.min, cmd.max, cmd.step);
         } else if (cmd.type == "set_control_value") {
             this.set_control_value(cmd.name, cmd.value, cmd.invoke_callback);
         } else if (cmd.type == "delete_control") {
             this.delete_control(cmd.name);
         } else if (cmd.type == "capture_image") {
-            let imgdata = this.capture_image();
+            let w = cmd.xres || 1920;
+            let h = cmd.yres || 1080;
```
ManifoldFR, line 1062: default resolution is 1080p

src/index.js
```diff
@@ -877,7 +877,7 @@ class Viewer {
             w = this.dom_element.offsetWidth;
         }
         if (h === undefined) {
-            h = window.innerHeight;
+            h = this.dom_element.offsetHeight;
```
rdeits, line 880: What's the intended effect of this change?

rdeits (approved):
Looks great, thanks!

src/index.js
rdeits: Ah, I see the references to this later. Ok, I'm happy with this.

Issue #114, opened 2022-04-19
set_control / set control_value : how to use callbacks


bbdb68:
Hello, I am using meshcat from c++ and this is really a nice tool.  

I try to use the set_control and set_control_value api, but I have an issue with callbacks : the callback (only `console.log('test value='+value);`  at the monent  is called when set_control is issued, but not when I call set_control_value (BTW the value is correctly set),  and also is it not called when I move the slider as a user.

Am I missing something ? thanks

(of course the next step would be to send back a message to the calling C++ code when a slider is modified)

RussTedrake:
Both of those should work; they are covered by the following test https://github.com/rdeits/meshcat/blob/master/test/controls.html#L46 (which I've just verified works fine on master).

Do you have a reproduction?

PR #115, opened 2022-04-21
Add UUID-based caching of geometry and materials

gizatt:
At present, even if a geometry or material is sent twice with the same UUID, there's no caching (at the parsing level, anyway -- maybe there's something deeper) of that declared-to-be-duplicated info. This enables UUID-based caching by creating a persistent `ExtensibleObjectLoader` that keeps a cache of UUIDs across all parsing calls (which is more than the vanilla `ObjectLoader` does). This change can be taken advantage of by clients (e.g. [meshcat-python PR #114](https://github.com/rdeits/meshcat-python/pull/114)) to save tons of bandwidth.


jwnimmer-tri:
I've been starting to consider some of these same improvements myself.

I'm curious if you've thought about how to keep the cache from growing without bound?  If I'm reading this code correctly, it seems like if I keep calling `set_object` over and over again on the same `path` with a new object each time, eventually the cache would become unreasonably large and run out of memory, with all of the cached-but-never-reused objects.

There's also the twin side of this -- the code in `dispose_recursive()` seems like it would dispose any loaded textures on the _first_ copy of the object we remove from the scene, leaving the other copies of the object without their textures any more?

jwnimmer-tri:
Closing this as incomplete / abandoned.

PR #129, opened 2022-09-17
Allow keyboard events to move sliders and click button controls.

RussTedrake:
My goal is to enable keyboard teleoperation through meshcat, even when I'm running it the Deepnote cloud service (which doesn't fully support ipywidgets).  This seemed like a particularly clean way to accomplish that; I've already been using sliders and buttons for teleop successfully in my course notebooks. This will make it much more ergonomic!

Actually the immediate motivation for coding it this weekend actually came from trying to convince a FIRST robotics team to use simulation. ;-)


src/index.js
```diff
@@ -1004,11 +1021,45 @@ class Viewer {
...
+                this.keydown_callbacks[keycode2] = [keydown_callback];
+              }
+            }
         } else {
             handler[name] = eval(callback);
             this.gui_controllers[name] = this.gui.add(handler, name);
             // The default layout for dat.GUI buttons is broken, with the button name artificially truncated at the same width that slider names are truncated.  We fix that here.
             this.gui_controllers[name].domElement.parentElement.querySelector('.property-name').style.width="100%";
+            if (keycode1 != undefined) {
+              let keydown_callback = {name: name, callback: eval(callback)};
```
rdeits, line 1056: We're now `eval`ing the callback twice on this code path and once above. Can you centralize that to remove the duplicate call (and in case we someday want to do something other that naively eval)?

src/index.js
```diff
@@ -1004,11 +1021,45 @@ class Viewer {
...
+                value = this.gui_controllers[name].getValue();
+                let new_value = Math.min(Math.max(value + step, min), max);
+                this.gui_controllers[name].setValue(new_value);
+              }};
+              if (keycode2 in this.keydown_callbacks) {
+                this.keydown_callbacks[keycode2].push(keydown_callback);
+              } else {
+                this.keydown_callbacks[keycode2] = [keydown_callback];
+              }
+            }
```
rdeits, line 1049: The two blocks here are identical except for the sign of the step. Make this a function instead?

rdeits (commented):
good idea, and good luck with the FIRST demo!

src/index.js
RussTedrake: Done.

src/index.js
RussTedrake: Done.

rdeits (approved):
:+1:

PR #130, opened 2022-09-20
Cleanup keypress handler if set_control overwrites an existing control

RussTedrake:
My Drake code reviewer grilled me about what happens when a second control of the same name is added. This change handles that program flow better; properly deleting the previous keypress handler.


RussTedrake:
thinking more, I actually like the ability to associate multiple keyboard events with the same button.  and the existing code handles that case properly.

PR #132, opened 2022-10-29
Switch from msgpack-lite to msgpack-javascript

RussTedrake:
See https://github.com/RobotLocomotion/drake/issues/17861 for a detailed description and analysis. We've seen approximately a 10x speed up in demarshalling approximately 30 MB of scene data.

This includes a small refactor of the handle_command to allow decode() to be called directly (since consumers of the Viewer might want to use the additionally defined msgpack types).

Also fixes a small bug with accumulating camera/light controllers.


src/index.js
```diff
@@ -9,6 +9,43 @@ import {STLLoader} from 'three/examples/jsm/loaders/STLLoader.js';
...
+  type: 0x16,
+  encode: (obj) => {
+    console.error("Uint32Array encode not implemented")
+    return null;
+  },
+  decode: (data) => {
+    const to_return = new Uint32Array(data.byteLength / 4);
+    let dataview = new DataView(data.buffer, data.byteOffset, data.byteLength);
+    for (let i = 0; i < to_return.length; i++) {
+      to_return[i] = dataview.getUint32(i * 4, true);  // true b.c. littleEndian
```
rdeits, line 26: It's still surprising that a JS loop over every byte of data could somehow be performant but :shrug: V8 is magic

rdeits (approved):
Cool! I'm glad there's a better option nowadays

PR #134, opened 2022-11-18
Fix for static_html logic in meshcat-python

RussTedrake:
It turns out that meshcat-python was using the
`handle_command_bytearray` entry point for the static html content. 
My PR #132 broke that (sorry!). This restores that interface and functionality.


rdeits (approved):
Thanks for catching this!

PR #137, opened 2022-12-07
Set properties recursively

petrikvladimir:
This PR allows modifying member variables of the object, examples:

```
obj._set_property('material.color.b', 1)  # change only the blue channel of the color
obj1._set_property('material.color', [0, 1, 0]) # change the color of the material
obj1._set_property('material.opacity', 0.5) # change the opacity of the material
```

How it works:
 - it splits strings based on dots to access internal variables, e.g., 'material.color.g' results in accessing variable 'g' of object `this.object.material.color`
 - if the accessed object is of type color and the value is an array, use set.RGB()
 - if the accessed object has a function set, use it to set the properties
 - otherwise, use the assignment operator to change the value

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -407,11 +407,18 @@ class SceneNode {
         } else if (property == "top_color" || property == "bottom_color") {
             this.object[property] = value.map((x) => x * 255);
         } else {
-            let v = this.object;
-            for (const p in property.split('.')) {
-                v = v[p];
+            function smart_set(obj, is, value) {
+                if (is.length !== 1) {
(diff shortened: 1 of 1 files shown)


jwnimmer-tri:
Thanks for your interest!

We recently did something similar in https://github.com/meshcat-dev/meshcat/pull/177.  Please take a look and see if it helps you out.

If there's still more to be done, please rebase this onto the latest master (or open a new pull request).

If not, please let us know and we can close this PR.  Thanks!

jwnimmer-tri:
Closing hoping that #177 was enough.  If not, feel free to re-open / open a new PR with a new take.

PR #138, opened 2022-12-14
Fix binary STL loading following switch from msgpack-lite to msgpack-javascript

wxmerkt:
Following the switch to msgpack-javascript from msgpack-lite (#132), we were no longer able to load binary STL meshes via meshcat-python: When decoding the message, it would leave the `decoded.geometries[i].data` as an object (with type and raw data) rather than the decoded array.

Specialising the decoder for Uint8Array as well resolves this. It's a bit of a mystery to me why it did not decode it out of the box, so any insights (or better solutions) much appreciated :-)


jwnimmer-tri:
Thanks for this.

The backstory is that `msgpack-lite` predefined a bunch of extension codecs (https://github.com/kawanet/msgpack-lite/tree/master#extension-types), but `msgpack-javascript` leaves it all up to us.

When this code switched its msgpack library (#132), only two extension types (`float32` and `uint32`) were retrofitted back in, even though [meshcat-python supports 4 extension types](https://github.com/meshcat-dev/meshcat-python/blob/785bc9d5ba6f8a8bb79ee8b25f523805946c1fbd/src/meshcat/geometry.py#L387-L397).  We're still missing `uint8` and `int32`.

I'll work on cleaning up the whole program, and making sure the test cases catch it.

PR #144, opened 2023-06-20
Accept embedded gltf files as _meshfile_objects

SeanCurtis-TRI:
This accepts embedded .gltf files as _meshfile_objects. It may also accept distributed gltf files as long as the meshcat server respects meshcat's requests for the missing .bin and image files.

This cannot be part of the ExtensibleObjectLoader (EOL). EOL inhertis from ObjectLoader where as GLTFLoader inherits from Loader. The workflow of ObjectLoader is incompatible with GLTFLoader's workflow. Fortunately, GLTFLoader produces a THREE.Group that we can simply stash in the scene.

To support PBR materials (reflective surfaces which reflect an environment map), we change the "background" gradient to support environment textures. This is largely an innocuous change. However, there is a distinct difference between the new background and the old.

In the old, the gradient was registered in *screen* space. Whether I was looking up or down, the top color was top of the *screen* and the bottom at the bottom of the *screen*.

The new environment map now use the "top" (+z infinitely far away) and bottom (-z infinitely far away) of the *world*. So, rotating the camera to look up sees the upper pole and looking down shows the lower pole.

- - -
This change is [Reviewable](https://reviewable.io/reviews/rdeits/meshcat/144)


SeanCurtis-TRI:
Here are a couple of screen grabs for what this enables (taken from a hacked version of Drake's [meshcat_manual_test.cc](https://github.com/RobotLocomotion/drake/blob/master/geometry/test/meshcat_manual_test.cc).

gltf_pr_images
Simple, abstract cube with a non-trivial material.

gltf_pr_images2
The canonical "damaged helmet" as featured on [Khronos's site](https://threejs.org/examples/?q=gltf#webgl_loader_gltf).

jwnimmer-tri:
\CC @pathammer in case you have any suggestions

src/index.js
```diff
@@ -998,22 +1006,49 @@ class Viewer {
...
-                if ((obj.geometry.attributes.normal === undefined) || obj.geometry.attributes.normal.count === 0) {
-                    obj.geometry.computeVertexNormals();
-                }
-            } else if (obj.type.includes("Camera")) {
-                this.set_camera(obj);
-                this.set_3d_pane_size();                
-            }
+        let configure_obj = (obj) => {
             obj.castShadow = true;
             obj.receiveShadow = true;
```
pathammer, line 1011: Do these overwrite the properties set on line 270 in ExtensibleObjectLoader::parseObject? It reads like they might be called second and overwrite.

src/index.js
pathammer, line 1: Good explanation for why this can't go in `ExtensibleObjectLoader`  :100: 

Do the OBJ json properties like [position, rotation, quaternion, scale, shadow.bias, shadow.radius, visible, frustumCulled, etc.] not apply to GLTF scene objects? If they do I would advise putting them in your new `configure_obj` arrow func.

pathammer, line 1: What about adding a GLTF test file that would serve as a good demo of the input format.

SeanCurtis-TRI: Yes, definitely. A glTF test. I just looked at the tests for the of the rmeshfile objects. I've now added the missing test.

SeanCurtis-TRI (commented):
Thanks for taking a look. I've responded to your comments in a variety of ways. Please take a look.

src/index.js
pathammer: 1. Most are Object3d properties (don't reconize userJson) - I think the docs were supposed to be [here](https://github.com/mrdoob/three.js/wiki/JSON-Object-Scene-format-4) but they aren't complete.
2. While we could apply the same properties to the top level gltf.scene object returned, that probably isn't helpful so I agree a wider brush is needed.   TODO containing clear note how it's unimplemented  sounds reasonable.

pathammer (approved):
I'm just a meshcat user with no authority here but for what it's worth LGTM

RussTedrake:
Thanks @pathammer !  Very helpful.

@SeanCurtis-TRI -- when I run the new test (thank you for adding it), the rendering looks fine by my browser java console warns
```
main.min.js:2 THREE.WebGLTextures: Unsupported texture encoding: srgb-linear
H @ main.min.js:2
k @ main.min.js:2
R @ main.min.js:2
Bo @ main.min.js:2
da.upload @ main.min.js:2
(anonymous) @ main.min.js:2
renderBufferDirect @ main.min.js:2
Oe @ main.min.js:2
De @ main.min.js:2
Ie @ main.min.js:2
render @ main.min.js:2
update @ main.min.js:2
fromEquirectangularTexture @ main.min.js:2
get @ main.min.js:2
render @ main.min.js:2
render @ main.min.js:2
render @ main.min.js:2
animate @ main.min.js:2
Viewer @ main.min.js:2
(anonymous) @ meshfile_object_gltf.html:13
```
Should I be worried?

RussTedrake (commented):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/rdeits/meshcat/144)* status: all files reviewed, 3 unresolved discussions (waiting on @pathammer and @SeanCurtis-TRI)

SeanCurtis-TRI (commented):
If you try it again, I think you can be less worried.

While I haven't tracked down the dependencies properly, basically, three.js is simultaneously declaring a particular attribute as deprecated, but *still* requiring it. So, I went ahead and patched the deprecated property back in, and it stops yelling. Go figure.

> *[Reviewable](https://reviewable.io/reviews/rdeits/meshcat/144#-:-NZPErb5BrCIPIGdwTqB:bht7oe)* status: all files reviewed, 3 unresolved discussions (waiting on @pathammer)

*[a discussion](https://reviewable.io/reviews/rdeits/meshcat/144#gh-1243089755:-NZPF5_ODw-r_yo4gcUs:b-896fix) (no related file):*
<details><summary><i>Previously, pathammer (Patrick Hammer) wrote…</i></summary><blockquote>

1. Most are Object3d properties (don't reconize userJson) - I think the docs were supposed to be [here](https://github.com/mrdoob/three.js/wiki/JSON-Object-Scene-format-4) but they aren't complete.
2. While we could apply the same properties to the top level gltf.scene object returned, that probably isn't helpful so I agree a wider brush is needed.   TODO containing clear note how it's unimplemented  sounds reasonable.
</blockquote></details>

Done.

___

RussTedrake (approved):
:lgtm:

> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/rdeits/meshcat/144#-:-NZRqqEeDPOv6S3EUDpV:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @pathammer)

SeanCurtis-TRI:
Thanks, @RussTedrake.

PR #147, opened 2023-07-19
glTF files now cast and receive shadows

SeanCurtis-TRI:
Previously, the cast/receive shadow properties were only set on the *root* of the glTF scene -- a three.js Group node. Because the glTF object can comprise an arbitrary hierarchy of geometry, each of the meshes in the hierarchy need to be individually configured, otherwise they won't participate in shadows.

Update the gltf test:
 - gltf now has the ability to cast shadows on itself.
 - Shadows are automatically enabled.

- - -
This change is [Reviewable](https://reviewable.io/reviews/rdeits/meshcat/147)


RussTedrake (approved):
:lgtm:

> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/rdeits/meshcat/147#-:-N_hiTFs7k_1jZpN2_Rx:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SeanCurtis-TRI)

PR #148, opened 2023-07-24
Extend "Background" to include an environment map

SeanCurtis-TRI:
Basic 2D images (jpg, png, etc.) can now be loaded as environment maps. This affects the illumination of objects in the scene for PBR materials.

Furthermore, we've added GUI controls for the default gradient colors so it can be set in session rather than strictly through the API. The addition of GUI controls means that the preferred path to Background is no longer simply `/Background`, but `/Background/<object>`. That said, the old spelling still works but prompts users to change their reference.

Finally, we've modified the meshfile_object.gltf.html test/example to include the environment map (it already has an object which can easily be made metallic to show reflection of the environment map). The page also gets an encoded environment map -- a custom image representing the classic Cornell box.

The basic logic is that if an environment map is defined, it is used for both background and material environment. However, there is an option for those who feel the rendered environment map is too distracting; it can be decoupled from the environment map and simply render the gradient -- the gradient will not illuminate materials in this case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/rdeits/meshcat/148)

PR #149, opened 2023-07-27
Fix setting background visibility vis set_property

SeanCurtis-TRI:
Introducing the environment map had introduced a bug in background visibility. Toggling the "Background" *folder (the GUI element) successfully toggled background visibility. But calling set_property() on it, didn't work.

This changes how the folder is defined so that it is still valid to call `set_property("\Background", "visible", bool)`.


SeanCurtis-TRI:
This fixes a bug introduced in #148.

PR #151, opened 2023-08-15
Make the control panel horizontally resizable

trowell-tri:
Enable the dat.GUI `resizable` flag, which allows resizing the control panel by dragging the left edge when the panel is open.


jwnimmer-tri:
We tested in in Drake via https://github.com/RobotLocomotion/drake/pull/20000 to good effect.

@RussTedrake please click the merge button.

RussTedrake:
I've tested it, too.  Nice!

PR #152, opened 2023-08-18
Fix illumination when background is invisible

SeanCurtis-TRI:
When the background is invisible, the environment map should be a bright white light field. It currently is not.

The cause for this was an initialization problem. The white gradients had the wrong texture mapping for flat and round. In flat you didn't notice because it was a badly rendered white square on a white background (it would've been visible if an image had been saved). The problem was more pronounced in the environment, because the wrong mapping keeps the white texture from illuminating the geometry.

It wasn't apparent in the *color* gradients, because when we reset any gradient we regenerate the gradients and put them in the right bins. But we never regenerate the white gradients so they stay wrong the whole time.

The fix is simply making sure that the textures get initialized properly.

The background.html test has been extended to articulate what should happen when the background is invisible so that the behavior is now locked to an observable test. Without the fix, marking the background as invisible would turn the box black.


RussTedrake:
Confirmed that background.html now looks better when the background is disabled.

PR #156, opened 2023-11-03
Allow registration of user callback invoked with each rendering

SeanCurtis-TRI:
Allows the introduction of arbitrary code to be executed each time the render pass is evaluated.

The render_callback.html exercises the new functionality.

- - -
This change is [Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156)


SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156)* status: 0 of 3 files reviewed, 1 unresolved discussion

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NiLnYPZ7yp7a_ZqWKaS:-NiLnYPZ7yp7a_ZqWKaT:bs1ri8d) (no related file):*
@jwnimmer-tri I'd like a second set of eyes on this.  I've made the test file (camera_callback.html) a test bed for the horrible quality of the camera matrix. I'm going to keep prodding at seeing if I'm simply doing a *bad* thing. But I thought I'd invite another pair of eyes.

When you load `camera_callback.html`, it should have sufficient in-browser documentation to let you know what's wrong and to recognize if it's fixed.

___

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156)* status: 0 of 3 files reviewed, all discussions resolved

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NiLnYPZ7yp7a_ZqWKaS:-Ni_IScDDWwiI6JC_-qw:b-krlu33) (no related file):*

@jwnimmer-tri I'd like a second set of eyes on this.  I've made the test file (camera_callback.html) a test bed for the horrible quality of the camera matrix. I'm going to keep prodding at seeing if I'm simply doing a *bad* thing. But I thought I'd invite another pair of eyes.

When you load `camera_callback.html`, it should have sufficient in-browser documentation to let you know what's wrong and to recognize if it's fixed.
</blockquote></details>

Nevermind. I had my epiphany on Sunday morning. I'm now getting good matrices all the time. Now I've got a different problem.

___

jwnimmer-tri:
Generally I would just remind you to take a step back and look at the big picture.  This is a yak pile that's pretty deep, and far away from your actual objective.  Lateral thinking might help you find less-general solutions that are closer at hand.

For example:
* Don't try to stream (push) the camera positions from the JS back to the C++.  Instead, offer C++ a way to poll for _any_ value from the JS side.  The model visualizer can manually poll for what it needs.
* Use PoseSliders for both the camera position and camera target (instead of JointSliders).
* Write a better default dmd.yaml file for the scene you'd like to interrogate.  Add multiple cameras for the multiple points of view you want to inspect.  Don't try to do anything interactive.

SeanCurtis-TRI (commented):
We had a race conditions. The new version is much simpler.

> offer C++ a way to poll for _any_ value

That's what the render-time callback allows for now.

> Use PoseSliders....better default dmd.yaml file.

I'm not quite sure what to make of these two comments. This PR doesn't provide the necessary context (I expect you're extrapolating from here to Drake but in a different direction than I did) and I'm going to wait until I push the Drake PR so you can see what that looks like. Then Drake-specific comments will have the appropriate context.

BTW For this solution, at the last second I had qualms about the callback mechanism as a way to inject arbitrary execution into the browser. But it's no worse than what we've been doing for getting slider values. So, meh.

> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-:-Ni_OtoC6_C_blMa94ol:b-tqskv7)* status: 0 of 3 files reviewed, all discussions resolved

SeanCurtis-TRI (commented):
I see what you're saying. This follows the 'get slider value" mechanism of reading values back from meshcat:

- Create a callback that grabs arbitrary state of the viewer, packages it in an arbitrary object, and then push it over the websocket.
- The C++ side then interprets messages that come back a very specific way.

This isn't the more generic mechanism of "get_property" that I call one time. It's not clear to me how to even achieve that as all communication is asynchronous. The request and response happen outside of the main thread so there's no way to get any particular value when you want it.

This is, however, a means to have a live stream of any value you care about updating at approximately same rate as the viewer display itself. From a user of `Meshcat`, it *looks* like you're simply querying and the back end implementation can evolve to make the illusion a reality.

To see this in action, see https://github.com/RobotLocomotion/drake/pull/20484.

> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-:-Ni_QsIa81-0Jv-zumao:b-9z3rta)* status: 0 of 3 files reviewed, all discussions resolved

jwnimmer-tri:
> This follows the 'get slider value" mechanism of reading values back from meshcat:

The difference is, the controls panel only sends a message when something _changes_.  This (as used by the Drake PR) seems to send an incredibly high rate of messages (from all connected browsers).

> It's not clear to me how to even achieve that as all communication is asynchronous.

See the prototype of image-fetching in https://github.com/RobotLocomotion/drake/issues/18912.  The `meshcat.cc` sends a request and then waits for a reply with a busy-loop.

jwnimmer-tri:
In any case, it probably make more sense to discuss this as part of the Drake PR, since that's the top-down view.  I'll bounce over to there next.

SeanCurtis-TRI (commented):
> incredibly high rate

I'm willing to bet that the peak rate is lower than the slider's peak rate. They trigger on the change event of the sliders. So, if you're slide back and forth, we're firing off events at a much higher rate than the render rate (that was certainly true for the change events on the camera control). Remember, that if things aren't getting dirty in the scene, there's no new rendering and there's no new camera position sent.

> ...busy loop.

I'd already dismissed that out of hand. Nothing like a GUI with busy loops.

Please take a look at the drake PR (when your platform reviewer duties are done). I feel that the end result is *incredibly* compelling. So, if you're at all susceptible to an "ends justifies the means" type of argument, the ability to manipulate the camera so easily and so smoothly should carry a significant amount of weight.

> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-:-Nia6iWaEj5T0HEmr2iK:b-ghtp88)* status: 0 of 3 files reviewed, all discussions resolved

jwnimmer-tri (changes requested):
> I'm willing to bet that the peak rate is lower than the slider's peak rate.

I did a little test (with firefox).  Moving the sliders quickly, and moving the camera quickly, both topped out at around 60 Hz of messages.  So there is no problem on this front.

(The message rate _does_ differ when everything is idle, but that's a question for the Drake code, not this code.)

---

I'm partway through writing up comments on the Drake PR, but this PR is in pretty fair shape already.  I haven't yet pulled it locally to run the unit test, but I'll still checkpoint a few inter-related documentation comments now to help unblock things.

> Reviewed 2 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-:-NifbmE176LSjqacT0wV:b-8h88a0)* status: 2 of 3 files reviewed, 5 unresolved discussions (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NifcDbtEst6vVD6farA:-NifcDbtEst6vVD6farB:b8o6jff) (no related file):*
Working

I still need to run the new test case locally.  (I'm planning to do that only after I've finished reviewing the Drake PR.)

___
___
*[`src/index.js` line 1032 at r2](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NifaFhI8QCJRYT7l_1t:-NifaFhI8QCJRYT7l_1u:b-misz2i) ([raw file](https://github.com/meshcat-dev/meshcat/blob/b48a04c7dba9e041e2f46476bd426981b59467b2/src/index.js#L1032)):*
> ```JavaScript
>         // Callback gets invoked with each rendering. User can define the
>         // callback by calling the `set_render_callback` command. Defaults to a
>         // no-op.
> ```

Adding inline documentation here is inconsistent with the pattern for the other nearby attributes.  I guess in C++ we're accustomed to having documentation a member field at its point of declaration, but it doesn't seem to be the case for a prototype-based language like JS (where we only have the point of initialization, not the point of declaration, available to us).

I don't see any clarity / readability problem of having `set_render_callback` be the only place where these details are documented.
___
*[`src/index.js` line 1286 at r2](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-Nifazr55WZtfEQ5DaE7:-Nifazr6CLQbr1O_6ewM:b-njixm9) ([raw file](https://github.com/meshcat-dev/meshcat/blob/b48a04c7dba9e041e2f46476bd426981b59467b2/src/index.js#L1286)):*
> ```JavaScript
>         // Sets the per-render use callback. `callback` should be a string that
>         // gets evaluated into an anonymous function (the string "null" will
>         // clear the callback).
> ```

I _suppose_ we can copy-paste the (forthcoming) API doc in the Readme into a JS comment within the function here if you feel strongly about it, but does seem to go against conventions of this project.  If you think we should follow a pattern of documenting the API inline for clarity, we should probably do the same across the whole project (maybe in another PR) so that the (new) convention would followed uniformly.

My gut feeling, though, is that a better plan would be to write good documentation in the Readme (with examples, spelling out any caveats, etc) and try not to duplicate any of that content here.

That's both the most similar to the current policy (so the least work), and to me seems like a pretty good habit on the merits, anyway.
___
*[`src/index.js` line 1525 at r2](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NieBR0g9tvRFm1wcGgj:-NieBR0g9tvRFm1wcGgk:b-bdw0or) ([raw file](https://github.com/meshcat-dev/meshcat/blob/b48a04c7dba9e041e2f46476bd426981b59467b2/src/index.js#L1525)):*
> ```JavaScript
>         } else if (cmd.type == "visualize_vr_controller"){
>             this.visualize_vr_controllers();
>         } else if (cmd.type == "set_render_callback") {
> ```

This seems to be adding to the official `handle_command` API of MeshCat.  Assuming so, then we'll need to document the new message specification in the reference manual, alongside the other commands:

https://github.com/meshcat-dev/meshcat#programmatic-api

(It might also be helpful to xref from the "Camera Control" chapter in the Readme to the new command documentation.)
___
*[`test/render_callback.html` line 3 at r2](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-Nif_sXp8gLCMGJjT94m:-Nif_sXp8gLCMGJjT94n:b-m8wd59) ([raw file](https://github.com/meshcat-dev/meshcat/blob/b48a04c7dba9e041e2f46476bd426981b59467b2/test/render_callback.html#L3)):*
> ```HTML
> <!DOCTYPE html>
> <html>
> 	<head>
> ```

nit Literal tabs.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156)* status: 2 of 3 files reviewed, 4 unresolved discussions (waiting on @jwnimmer-tri)

___
*[`src/index.js` line 1032 at r2](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NifaFhI8QCJRYT7l_1t:-Nig7yRf2ksID7RJEs96:b-896fix) ([raw file](https://github.com/meshcat-dev/meshcat/blob/b48a04c7dba9e041e2f46476bd426981b59467b2/src/index.js#L1032)):*

Adding inline documentation here is inconsistent with the pattern for the other nearby attributes.  I guess in C++ we're accustomed to having documentation a member field at its point of declaration, but it doesn't seem to be the case for a prototype-based language like JS (where we only have the point of initialization, not the point of declaration, available to us).

I don't see any clarity / readability problem of having `set_render_callback` be the only place where these details are documented.
</blockquote></details>

Done.
___
*[`src/index.js` line 1286 at r2](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-Nifazr55WZtfEQ5DaE7:-Nig7rp3-asAn4nrsBAf:bb6un1j) ([raw file](https://github.com/meshcat-dev/meshcat/blob/b48a04c7dba9e041e2f46476bd426981b59467b2/src/index.js#L1286)):*

I _suppose_ we can copy-paste the (forthcoming) API doc in the Readme into a JS comment within the function here if you feel strongly about it, but does seem to go against conventions of this project.  If you think we should follow a pattern of documenting the API inline for clarity, we should probably do the same across the whole project (maybe in another PR) so that the (new) convention would followed uniformly.

My gut feeling, though, is that a better plan would be to write good documentation in the Readme (with examples, spelling out any caveats, etc) and try not to duplicate any of that content here.

That's both the most similar to the current policy (so the least work), and to me seems like a pretty good habit on the merits, anyway.
</blockquote></details>

Done
___
*[`src/index.js` line 1525 at r2](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NieBR0g9tvRFm1wcGgj:-Nig7WO2-nXyDtLTWHUJ:b-k1dtdt) ([raw file](https://github.com/meshcat-dev/meshcat/blob/b48a04c7dba9e041e2f46476bd426981b59467b2/src/index.js#L1525)):*

This seems to be adding to the official `handle_command` API of MeshCat.  Assuming so, then we'll need to document the new message specification in the reference manual, alongside the other commands:

https://github.com/meshcat-dev/meshcat#programmatic-api

(It might also be helpful to xref from the "Camera Control" chapter in the Readme to the new command documentation.)
</blockquote></details>

Done

It's worth noting that there are a number of commands that have not been added. A little follow up clean up would probably be a good thing.

jwnimmer-tri (commented):
> Reviewed 2 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @SeanCurtis-TRI)

___
*[`Readme.md` line 285 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NigCl0FDrp2jTvwEpfU:-NigCl0GA0CYjhvgfYy-:b-czyinq) ([raw file](https://github.com/meshcat-dev/meshcat/blob/9775c75f0e5a88a1769d8f87108ddf5d899188eb/Readme.md#L285)):*
> ```Markdown
>                 can be used to execute arbitrary code per rendered frame.
>                 <br><br>
>                 The callback is passed the `viewer` doing the rendering and the
> ```

BTW I just realized -- I wonder if we're being weird by passing the `viewer` as a hard-coded argument.  I have a slight intuition that the "JavaScript Way" would be to have a zero-argument callback, so it's the user's job to prepare a closure with whatever captures they need (the `viewer` or otherwise).  If they don't actually need the viewer, they would need to wrangle with throwing away the unnecessary argument.
___
*[`src/index.js` line 1525 at r2](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NieBR0g9tvRFm1wcGgj:-NigBvx389vIJZQUgZF9:bt1kmir) ([raw file](https://github.com/meshcat-dev/meshcat/blob/b48a04c7dba9e041e2f46476bd426981b59467b2/src/index.js#L1525)):*

Done

It's worth noting that there are a number of commands that have not been added. A little follow up clean up would probably be a good thing.
</blockquote></details>

Yeah, I noticed that also.  Even some of what is there in the Readme is incorrect.

jwnimmer-tri (changes requested):
:lgtm:

> Reviewed 1 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-:-NijFI9SCI7PR59Cx5NZ:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NifcDbtEst6vVD6farA:-NijFLBP3u2MN_PGbOx7:bb6un1j) (no related file):*

Working

I still need to run the new test case locally.  (I'm planning to do that only after I've finished reviewing the Drake PR.)
</blockquote></details>

Done

___
___
*[`test/render_callback.html` line 77 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NijFAdt7jcAFVrzQ3JS:-NijFAdt7jcAFVrzQ3JT:b7j9by7) ([raw file](https://github.com/meshcat-dev/meshcat/blob/9775c75f0e5a88a1769d8f87108ddf5d899188eb/test/render_callback.html#L77)):*
> ```HTML
>                     type: "set_render_callback",
>                     callback: callback_state == 0 ?
>                         null :
> ```

The Readme says that the callback should always be a string, so we want the string `"null"` here I think, not a literal `null`?

SeanCurtis-TRI (commented):
I've added another PR. The other PR is what this was originally intended to be. It is purely the callback for when the camera pose changes (I've figured out how to stabilize the value). Take a look at #157.

> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-:-Nikr-WL4jTJeuee0AXe:bszmngz)* status: 1 of 4 files reviewed, 2 unresolved discussions (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-Niko0vjCMqzNUEAYhTM:-Niko0vjCMqzNUEAYhTN:b-oitezj) (no related file):*
Blocking for design discussion.

___
___
*[`Readme.md` line 285 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NigCl0FDrp2jTvwEpfU:-NikXSML1x4ECcujLBoF:b-hhwfa3) ([raw file](https://github.com/meshcat-dev/meshcat/blob/9775c75f0e5a88a1769d8f87108ddf5d899188eb/Readme.md#L285)):*

BTW I just realized -- I wonder if we're being weird by passing the `viewer` as a hard-coded argument.  I have a slight intuition that the "JavaScript Way" would be to have a zero-argument callback, so it's the user's job to prepare a closure with whatever captures they need (the `viewer` or otherwise).  If they don't actually need the viewer, they would need to wrangle with throwing away the unnecessary argument.
</blockquote></details>

The slider callback explicitly refers to `this`. That kind of practice makes assumptions about the context in which the string gets evaluated. It seemed to make the callback clearer to simply pass the `viewer` that the callback should have access to and then the person defining the context didn't have to worry about the context in which it gets evaluated. Everything it needs is provided as an argument.

Certainly, most of the javascript callbacks I've looked at in three.js define parameters. The nice thing is that just because meshcat promises to pass viewer doesn't mean you have to define a callback that declares the parameter. So, the user wouldn't need to "wrangle with throwing [things] away". They can happily pretend it never happened.

So, I think we're ok.
___
*[`test/render_callback.html` line 77 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NijFAdt7jcAFVrzQ3JS:-Niknvkh93Ha152Z0KZN:b-oqcz30) ([raw file](https://github.com/meshcat-dev/meshcat/blob/9775c75f0e5a88a1769d8f87108ddf5d899188eb/test/render_callback.html#L77)):*

The Readme says that the callback should always be a string, so we want the string `"null"` here I think, not a literal `null`?
</blockquote></details>

Done

And that revealed a bug. Good eye.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-Niko0vjCMqzNUEAYhTM:-NipVC7z8shgmtEfRXji:bphbmxw) (no related file):*

Blocking for design discussion.
</blockquote></details>

I think we're good on having the render callback on the meshcat side. So, I'm going to stop blocking here.

___
___
*[`test/render_callback.html` line 77 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NijFAdt7jcAFVrzQ3JS:-NipVA2XApuNEpWRaFsX:b3eqhms) ([raw file](https://github.com/meshcat-dev/meshcat/blob/9775c75f0e5a88a1769d8f87108ddf5d899188eb/test/render_callback.html#L77)):*

Done

And that revealed a bug. Good eye.
</blockquote></details>

This is waiting on you.

jwnimmer-tri (changes requested):
> Reviewed 3 of 3 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156)* status: all files reviewed, 2 unresolved discussions (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NipVxjqBlcqOeRBLiP2:-NipVxjqBlcqOeRBLiP3:b-m81hye) (no related file):*
Working

Blocking merge until the Drake PR is sufficiently approved / finished.

(Also we'll need a platform reviewer on this PR at some point.)

___
___
*[`test/render_callback.html` line 65 at r5](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NipVjUBEXOejAnrcgVt:-NipVjUBEXOejAnrcgVu:b-9ixtoo) ([raw file](https://github.com/meshcat-dev/meshcat/blob/533779e3bc392ac51b4a3a9a3b74ac81df3ae2f0/test/render_callback.html#L65)):*
> ```HTML
>                 "Callback prints camera pose to console each time the scene " +
>                 "is rendered. Try toggling the visibility of lights or the " +
>                 "grid. Each time it toggles, the camera matrix will be printed."
> ```

Working

I still need to try running this new advice locally.

jwnimmer-tri (changes requested):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156)* status: all files reviewed, 3 unresolved discussions (waiting on @SeanCurtis-TRI)

___
*[`Readme.md` line 285 at r5](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NjEQ2H15nOZtqq03eF3:-NjEQ2H15nOZtqq03eF4:b-cxz30x) ([raw file](https://github.com/meshcat-dev/meshcat/blob/533779e3bc392ac51b4a3a9a3b74ac81df3ae2f0/Readme.md#L285)):*
> ```Markdown
>                 can be used to execute arbitrary code per rendered frame.
>                 <br><br>
>                 The callback is passed the `viewer` doing the rendering and the
> ```

See https://github.com/jwnimmer-tri/drake/commits/20484-send-only-changed.

There is _no reason_ to pass a `viewer` argument.  Even with a string lambda, both `viewer` and `this` are in scope already.  None of the other callbacks (e.g., buttons) have an argument.

SeanCurtis-TRI:
*[`Readme.md` line 285 at r5](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NjEQ2H15nOZtqq03eF3:-NjJl12XCqdHqY1E0Dl9:b-wjsu2l) ([raw file](https://github.com/meshcat-dev/meshcat/blob/533779e3bc392ac51b4a3a9a3b74ac81df3ae2f0/Readme.md#L285)):*

See https://github.com/jwnimmer-tri/drake/commits/20484-send-only-changed.

There is _no reason_ to pass a `viewer` argument.  Even with a string lambda, both `viewer` and `this` are in scope already.  None of the other callbacks (e.g., buttons) have an argument.
</blockquote></details>

So, essentially, we're promising that the viewer is available as `this` which puts constraints on when and how the javascript string gets evaluated. It seems so much *saner* to just pass the viewer. It's an easier promise to check.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @jwnimmer-tri)

___
*[`Readme.md` line 285 at r5](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NjEQ2H15nOZtqq03eF3:-NjK02GZ8jnVN8kQC0Gt:b-ua0npf) ([raw file](https://github.com/meshcat-dev/meshcat/blob/533779e3bc392ac51b4a3a9a3b74ac81df3ae2f0/Readme.md#L285)):*

So, essentially, we're promising that the viewer is available as `this` which puts constraints on when and how the javascript string gets evaluated. It seems so much *saner* to just pass the viewer. It's an easier promise to check.
</blockquote></details>

I yield.

jwnimmer-tri (approved):
> Reviewed 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NipVxjqBlcqOeRBLiP2:-NjK7h8Q1NvmAHa85ssx:bwaguu9) (no related file):*

Working

Blocking merge until the Drake PR is sufficiently approved / finished.

(Also we'll need a platform reviewer on this PR at some point.)
</blockquote></details>

I think this is OK for platform review now.  The Drake PR still has a couple of open discussions about testing, but overall it's pretty close.

___
___
*[`test/render_callback.html` line 65 at r5](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NipVjUBEXOejAnrcgVt:-NjK7cag7mEPd3xqOUvY:bb6un1j) ([raw file](https://github.com/meshcat-dev/meshcat/blob/533779e3bc392ac51b4a3a9a3b74ac81df3ae2f0/test/render_callback.html#L65)):*

Working

I still need to try running this new advice locally.
</blockquote></details>

Done

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @EricCousineau-TRI)

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NipVxjqBlcqOeRBLiP2:-NjKHMS9BpwQssDyIftQ:bed7lxu) (no related file):*

I think this is OK for platform review now.  The Drake PR still has a couple of open discussions about testing, but overall it's pretty close.
</blockquote></details>

+@EricCousineau-TRI too late for a Wednesday review?

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156)* status: all files reviewed, 1 unresolved discussion (waiting on @EricCousineau-TRI and @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NipVxjqBlcqOeRBLiP2:-NjKjwMA19gsgUhn_sSi:bg74q4) (no related file):*

+@EricCousineau-TRI too late for a Wednesday review?
</blockquote></details>

He's on call for Thursday, so hopefully the timing works out OK!

___

EricCousineau-TRI (approved):
:lgtm:, nice!
Did local testing, confirmed behavior is as expected.

> Reviewed 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-:-NjOHZfQ-d8NaU6_H28-:bu057is)* status: all files reviewed, 3 unresolved discussions (waiting on @SeanCurtis-TRI)

___
*[`Readme.md` line 292 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NjOFm1s2OmIvyRaHkyK:-NjOFm1s2OmIvyRaHkyL:bo6j0ys) ([raw file](https://github.com/meshcat-dev/meshcat/blob/8c49b8aaa227469e3ad3d327c2ed54228ef73946/Readme.md#L292)):*
> ```Markdown
>             <pre>
> {
>     "type": "set_render_callback",
> ```

nit It was not immediately clear that this was example.

Consider mentioning that? e.g. "The following is an example callback that dispatches the camera's pose in the world to the websocket connection.
___
*[`test/render_callback.html` line 119 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-NjOG8Gt1i-aiK14bYSm:-NjOG8Gt1i-aiK14bYSn:bkkajxz) ([raw file](https://github.com/meshcat-dev/meshcat/blob/8c49b8aaa227469e3ad3d327c2ed54228ef73946/test/render_callback.html#L119)):*
> ```HTML
>         </style>
>         <script id="embedded-json"></script>
> 	</body>
> ```

nit Whitespace (tab character?)

SeanCurtis-TRI (commented):
Comments all addressed.

> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156#-:-NkMnA1h1DC-bYLvkeb0:b-8s2m2n)* status: 2 of 4 files reviewed, all discussions resolved (waiting on @EricCousineau-TRI and @jwnimmer-tri)

jwnimmer-tri (approved):
> Reviewed 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/156)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SeanCurtis-TRI)

PR #157, opened 2023-11-08
Allow registration of user callback invoked with change in camera pose

SeanCurtis-TRI:
Allows the introduction of arbitrary code to be executed each time the the camer pose chages.

The camera_callback.html exercises the new functionality.

- - -
This change is [Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/157)


SeanCurtis-TRI (commented):
cc @jwnimmer-tri 

> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/157#-:-NiksMalENQTLSFEfN7W:bv9d4mg)* status: 0 of 4 files reviewed, 1 unresolved discussion

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/157#-Niks3Ot4suW8iHsjok3:-Niks3Ot4suW8iHsjok4:bui2aci) (no related file):*
Note: this still isn't the alternative proposal where it's fully embedded in meshcat.cc, but it's compatible with that. It has the unique benefit of getting new camera poses only when the camera changes. Achieving this requires some special incantations that it would be unreasonable for the end user to get right.

Technically, we could generalize this such that we create a "register event listenener" command where the command specifies the callback function and specifies the object and the event to which it should respond. That would open up a number of paths allowing the server side to decide the cadence of the callback.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/157)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/157#-Niks3Ot4suW8iHsjok3:-Nil_bHb5l7vnKSrd8_H:bz4u13k) (no related file):*

Note: this still isn't the alternative proposal where it's fully embedded in meshcat.cc, but it's compatible with that. It has the unique benefit of getting new camera poses only when the camera changes. Achieving this requires some special incantations that it would be unreasonable for the end user to get right.

Technically, we could generalize this such that we create a "register event listenener" command where the command specifies the callback function and specifies the object and the event to which it should respond. That would open up a number of paths allowing the server side to decide the cadence of the callback.
</blockquote></details>

This is consistent with discussions in the design document where MeshCat just invokes a callback and it's up to Drake to decide what to do about it (like sending a message).

___

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/157)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/157#-Niks3Ot4suW8iHsjok3:-NiobeKQ0IA9ntIioX1u:b-xdab8e) (no related file):*

This is consistent with discussions in the design document where MeshCat just invokes a callback and it's up to Drake to decide what to do about it (like sending a message).
</blockquote></details>

Consistent, yes. But it's distinct.

An argument in *favor* of this kind of thing:

- The fact that it is triggered by a camera change event means that you don't have to worry about it firing off for other reasons. It offers a *targeted* signal and response. Arguably requiring less filtering work on the part of the "user".
- It allows for a distinct camera callback from a render callback. They need not be explicitly coordinated. Both can cohabit.

Arguments *against* this kind of thing:

- Responding to the event could lead to a *higher* signal rate during direct manipulation than simply using the render callback. Do we really need to respond to camera changes faster than we render?

Both this and the render callback suffer from the idea that there is a *single* render callback registered with the viewer. So, if different components of the user code want different behaviors based on events/rendering/whatever, they have to explicitly coordinate. That seems clunky.

Perhaps what *should* be happening is that `Viewer` should store a dictionary of callbacks. When you register one, you're obliged to provide a *name* for the callback. The name is used to remove the callback and the name is used to differentiate independent requests to add a callback. And, in the case of two callbacks being registered with the same name, it can detect that and *do* something (I'm not yet clear on what the semantics of this would be).

___

SeanCurtis-TRI:

I agree with everything you wrote, up until to this point:

> ... they have to explicitly coordinate. That seems clunky.

It's not clunky.  If the user needs multi-dispatch, then can trivially set up their own dictionary with whatever semantics they want.

Having them screw around with our particular choice of semantics for callback "naming" is _more_ clunky.

---

In terms of whether we implement a render callback or a camera motion callback, I don't mind either way.  It sounds like in either case, it's possible that the Drake handler will need to do some rate-limiting anyway (depending on how high the "higher signal rate" is).
</blockquote></details>

I'm content with deferring this PR for #156. This is halfway between the render callback and what I would actually advocate if we did this. So, I'm simply going to close it. It can always be revisited at some tie in the future.

SeanCurtis-TRI:
Our purposes are served with the more general render callback (#156). So, we'll close this until we have some proven value.

PR #171, opened 2024-02-04
Fix loading glTF meshes with unbundled resources

jwnimmer-tri:
With Meshcat, the word 'path' can mean two different kinds of thing:

- A designator for a SceneNode, i.e., the names of elements to descend through in the scene tree (either as an array of names as seen in the JS code, or a slash-separated string as seen in the a message).

- An attribute of a FileLoader that specifies what URL prefix to use when looking up relative filenames as assets.

When passing 'path' to GLTFLoader.parse, the argument is of the second kind (a URL) but we were giving it a value of the first kind.

This didn't matter when the glTF only contains embedded assets, but causes trouble when it uses relative URIs: they end up being polluted by the SceneTree path and the web server has no idea what files the loader is asking for.


jwnimmer-tri (commented):
+@SeanCurtis-TRI for both reviews, please.

> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/171#-:-NpmLgZZ1bQCbG9xqmoo:b-fqjuvi)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @SeanCurtis-TRI)

PR #174, opened 2024-02-07
Fix DAE texture loading in _meshfile_object

jwnimmer-tri:


jwnimmer-tri:
This was a little demo to help #170.  I'll close it in lieu of that PR.

PR #177, opened 2024-02-22
Viewer.set_property can reference nested properties

### LINKED ISSUE #168: set_property on nested properties
Sometimes, a user might want to change a property that is nested within the scene_tree's object.

For example:

```js
{
  "type": "set_property",
  "path": "/Lights/PointLightNegativeX/<object>",
  "property": "shadow.radius",
  "value": 1.15
}
```

If the user attempts this today, Meshcat will do `scene_node.object["shadow.radius"] = 1.15`.

Instead, we want it to do `scene_node.object["shadow"]["radius"] = 1.15`.

SeanCurtis-TRI:
This comes with more feedback about attempts to set properties that don't exist (or can't be set).

Resolves #168


SeanCurtis-TRI (commented):
+@jwnimmer-tri for feature review, please.

> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-:-NrIPkQ8D4NKPGvTId5s:b-qf68jg)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @jwnimmer-tri)

jwnimmer-tri (commented):
FYI I don't have a lot of time to look at this today.  I can look tonight or first thing tomorrow morning.

> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-:-NraFWmvBLaCTG7xbvlw:bwbuwvf)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NraFVI1Drez_ApUpOGL:-NraFVI248BZNQX41fbB:b67m1vb) (no related file):*
Please add README.md API specification for the revised semantics.

___

jwnimmer-tri (commented):
This seems plausible on first skim, but I want us to be very careful here so I'm going to wait for the README specs before I carefully comb through the code.

> Reviewed 1 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-:-Nrbc0RsEDoh7DIuWm65:b-j061i8)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @SeanCurtis-TRI)

___
*[`src/index.js` line 711 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrbcR3t0GU2PSt-RcBI:-NrbcR3t0GU2PSt-RcBJ:bg8iuxc) ([raw file](https://github.com/meshcat-dev/meshcat/blob/3e4d168f1e04255bec1423e9f2e50cdc2de2894f/src/index.js#L711)):*
> ```JavaScript
>                 // property *name*. Furthermore, that property can be accessed
>                 // either as a["1"] or a[1].
>                 parent[child] = {};
> ```

BTW Per the issue thread, I'm inclined to leave this out until we understand some need for it.
___
*[`test/nested_properties.html` line 82 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrbOEMS7_wHGUef5BM6:-NrbOEMS7_wHGUef5BM7:b-ovfwwa) ([raw file](https://github.com/meshcat-dev/meshcat/blob/3e4d168f1e04255bec1423e9f2e50cdc2de2894f/test/nested_properties.html#L82)):*
> ```HTML
>             }
>         });
>         const path = "/meshcat/mesh_test";
> ```

nit Should be the path be defined one stanza prior (line 61) to reduce copy-paste?

SeanCurtis-TRI (commented):
I've added to the README and tweaked the test accordingly.

> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-:-NrffYOECB2wCTmvQG0f:b9mrhkk)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[`src/index.js` line 711 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrbcR3t0GU2PSt-RcBI:-NrffgPe7z7AD01dAf3n:b-jsayc4) ([raw file](https://github.com/meshcat-dev/meshcat/blob/3e4d168f1e04255bec1423e9f2e50cdc2de2894f/src/index.js#L711)):*

BTW Per the issue thread, I'm inclined to leave this out until we understand some need for it.
</blockquote></details>

See the note in readme.md. Philosophically, it's currently justified as paralleling the behavior of specifying a non-existent path. If we want to contrast it instead of parallel it, I'm good with that. I figured for my first pass on the README, I'd document what *is* and then we can change it in a unified manner.

jwnimmer-tri (commented):
> Reviewed 1 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @SeanCurtis-TRI)

___
*[`src/index.js` line 711 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrbcR3t0GU2PSt-RcBI:-NrfhvKqA9us8NWDxCj3:b4aizgs) ([raw file](https://github.com/meshcat-dev/meshcat/blob/3e4d168f1e04255bec1423e9f2e50cdc2de2894f/src/index.js#L711)):*
> Philosophically, it's currently justified as paralleling the behavior of specifying a non-existent path.

The difference is that paths are freeform, and it's expected that the user will choose their own convention for what paths they want to add into their scene.  Thus, creating them out of thin air when cited makes sense, because there is no wrong answer.

Properties are not freeform, they are whatever the threejs object knows about.  There is such a thing as a wrong answer.  When the user has a typo in the property name, I think we need to tell them about it (and it seems like we do that now, thankfully, in this PR).  Being silent about it was a real problem, historically.

Still, I don't see any reason why we want to add the dummy object as a JS property that threejs will never use.  I don't think "by analogy to paths" holds water, given the different context.

> If we want to contrast it instead of parallel it, I'm good with that.

As always I'm open to use cases where adding the JS object is helpful, but failing that I don't see any use in adding new objects.

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[`src/index.js` line 711 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrbcR3t0GU2PSt-RcBI:-NrfmtQ0BpQjUSdXRZ-7:bxdkozx) ([raw file](https://github.com/meshcat-dev/meshcat/blob/3e4d168f1e04255bec1423e9f2e50cdc2de2894f/src/index.js#L711)):*

> Philosophically, it's currently justified as paralleling the behavior of specifying a non-existent path.

The difference is that paths are freeform, and it's expected that the user will choose their own convention for what paths they want to add into their scene.  Thus, creating them out of thin air when cited makes sense, because there is no wrong answer.

Properties are not freeform, they are whatever the threejs object knows about.  There is such a thing as a wrong answer.  When the user has a typo in the property name, I think we need to tell them about it (and it seems like we do that now, thankfully, in this PR).  Being silent about it was a real problem, historically.

Still, I don't see any reason why we want to add the dummy object as a JS property that threejs will never use.  I don't think "by analogy to paths" holds water, given the different context.

> If we want to contrast it instead of parallel it, I'm good with that.

As always I'm open to use cases where adding the JS object is helpful, but failing that I don't see any use in adding new objects.
</blockquote></details>

This is where I think `meshcat` may be a bit crazy. I have `set_object()` which is the most logical way to introduce objects at a path. `set_property()` allows me to do so as well. However, what you're proposing is that while I can use `set_property()` to create an object, I can't subsequently set any property on it until I've called `set_object()` anyways.

I appreciate your point, Rhetorically, I'd focus less on the idea of "freeform" than I would on ownership. meshcat owns the scene graph (and the corresponding path), but properties are only set on Three.js objects.

Still, I'll go ahead and change it to make property setting a strict operation. Stay tuned...

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-Nrg6o0c9aNtvOJqCr0b:-Nrg6o0c9aNtvOJqCr0c:b-eoof0z) (no related file):*
BTW In Drake, `Meshcat::PublishRecording()` has an interesting comment about the fact that you can create an animation and publish it, but might not see anything. The comment is clearly a bit opaque to me, but I think the logic behind it is that I can set all of the properties animations before I've even set any objects.  THat seems a silly thing to do, but I bet that this PR would make that act a lot noisier in the console.

___
___
*[`src/index.js` line 711 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrbcR3t0GU2PSt-RcBI:-Nrg6HS0EX5ClPZ9Xppe:b-xhz5gd) ([raw file](https://github.com/meshcat-dev/meshcat/blob/3e4d168f1e04255bec1423e9f2e50cdc2de2894f/src/index.js#L711)):*

> However, what you're proposing is that while I can use `set_property()` to create an object, I can't subsequently set any property on it until I've called `set_object()` anyways.

Maybe I'm mistaken, but aren't there some properties created as soon as I add the SceneNode?  Like `"visible"`?  So it's not strictly true that I can't effectively set_property until after the set_object.

Any anyway, if I set a property ahead of time and then load an object atop it, it would stop all over my old property anyway right?

Maybe you should phone me, in case we're just talking past each other.
</blockquote></details>

Oh, no elaboration necessary. I've fallen into my classic trap where I'm not arguing a position, I'm arguing an argument. :) I have no problems with *not* creating properties. I just think the nature of `meshcat` itself weakens any definitive arguments for why one argument is definitively right and the other wrong.

PTAL.

jwnimmer-tri (commented):
The general feeling I came away with is that the JS code could be a lot more crisp and encapsulated.

> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-:-Nrkz1Uu9nF0SHbRTcPP:b-fc6shb)* status: all files reviewed, 5 unresolved discussions (waiting on @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-Nrh1QID-vJcp0mwWyL_:-Nrh1QIE-41U91DbRnf6:burm9aa) (no related file):*
The original call to action #168 had an example of "shadow.radius".  Have you hero-tested that an example like that works correctly now?

___
___
*[`src/index.js` line 664 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-Nrh04mSBlF4Rl1fPDi9:-Nrh04mSBlF4Rl1fPDiA:bu8ck69) ([raw file](https://github.com/meshcat-dev/meshcat/blob/3fb3b591bafa759b595587583978b59803a120f7/src/index.js#L664)):*
> ```JavaScript
>             this.object[property] = new dat.color.Color(value.map((x) => x * 255));
>         } else {
>             var property_sequence = this.split_property(property);
> ```

BTW I think the code would end up a lot more clear if the "split_property" returned two distinct things -- the (possibly empty) sequence of object navigation, and then then actual property name.  The slicing stuff inline here took me a little bit of time to grok.

Or after having read the whole patch a few more times, I'd instead suggest having two different local variables for the "object navigation" part of the "property name" part, and just compute them directly (inlining the helper function).
___
*[`src/index.js` line 709 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrkukvB9jyv80QqunOa:-NrkukvC8B05cOO9vvD0:b-pf3vws) ([raw file](https://github.com/meshcat-dev/meshcat/blob/3fb3b591bafa759b595587583978b59803a120f7/src/index.js#L709)):*
> ```JavaScript
>                 // Note: even if the next property name in the property path was
>                 // an index (e.g., a[1] became a.1), we can get away with adding
>                 // an object, because the index will simply be treated as a
> ```

nit "adding an object" seems like a vestigial comment?
___
*[`src/index.js` line 715 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrkwAaiCWSQxGHvKhFY:-NrkwAaiCWSQxGHvKhFZ:b41w1j1) ([raw file](https://github.com/meshcat-dev/meshcat/blob/3fb3b591bafa759b595587583978b59803a120f7/src/index.js#L715)):*
> ```JavaScript
>             }
>         } catch (error) {
>             console.error("When setting a chained property, the intermediate " +
> ```

BTW I'm not sure I buy the code cost from the two different log clauses.  The log severity (error vs warn) does not offer any meaningful benefit to our users.  Gut feeling, a simpler implement along these lines would be easier to live with, long-term:

```
try
  if (child in parent)
    return parent[child]
catch
  console error
  return null
```

Having it be shorter and simpler would also let us nix the function and move it back inline at the call site, so we don't need to jump around on the page and everything is better-encapsulated.

If anything, my gut feeling is that adding a single `set_property_generic` helper function (with no other helper functions) would be a better separation of concerns than keeping the logic inside of `set_property`.
___
*[`test/chained_properties.html` line 117 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-Nrkrk-b4KoMooScWtkD:-Nrkrk-b4KoMooScWtkE:bb7phvt) ([raw file](https://github.com/meshcat-dev/meshcat/blob/3fb3b591bafa759b595587583978b59803a120f7/test/chained_properties.html#L117)):*
> ```HTML
> 
>         console.info("Setting a *chained* property. The intermediate property " +
>                      "doesn't exist. We should get *one* warnings about " +
> ```

typo



_Suggestion:_
```HTML
warning
```

SeanCurtis-TRI (commented):
You've expressed a strong preference to see the three methods all become a single block. I'm not persuaded. But I'm not done yet. Part of the compression has to do with message logging. I've made an effort to argue in favor of errors and warnings. Before I erase the distinction, I wanted to see if you found the arguments convincing.

> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-:-NrlzoJZ4tB-y89aIc0f:b54h9gv)* status: all files reviewed, 1 unresolved discussion (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-Nrh1QID-vJcp0mwWyL_:-NrlYr6w9MzcG9A_VC2Q:bmnaqwg) (no related file):*

The original call to action #168 had an example of "shadow.radius".  Have you hero-tested that an example like that works correctly now?
</blockquote></details>

The call to `material.opacity` covers that.

___
___
*[`src/index.js` line 664 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-Nrh04mSBlF4Rl1fPDi9:-NrlzfDgEvFa2v4qwpWh:b90wb5e) ([raw file](https://github.com/meshcat-dev/meshcat/blob/3fb3b591bafa759b595587583978b59803a120f7/src/index.js#L664)):*

BTW I think the code would end up a lot more clear if the "split_property" returned two distinct things -- the (possibly empty) sequence of object navigation, and then then actual property name.  The slicing stuff inline here took me a little bit of time to grok.

Or after having read the whole patch a few more times, I'd instead suggest having two different local variables for the "object navigation" part of the "property name" part, and just compute them directly (inlining the helper function).
</blockquote></details>

I went ahead and returned the tuple. 
___
*[`src/index.js` line 715 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrkwAaiCWSQxGHvKhFY:-Nrl_Yo7ECyHD25IydR-:bjbslem) ([raw file](https://github.com/meshcat-dev/meshcat/blob/3fb3b591bafa759b595587583978b59803a120f7/src/index.js#L715)):*

BTW I'm not sure I buy the code cost from the two different log clauses.  The log severity (error vs warn) does not offer any meaningful benefit to our users.  Gut feeling, a simpler implement along these lines would be easier to live with, long-term:

```
try
  if (child in parent)
    return parent[child]
catch
  console error
  return null
```

Having it be shorter and simpler would also let us nix the function and move it back inline at the call site, so we don't need to jump around on the page and everything is better-encapsulated.

If anything, my gut feeling is that adding a single `set_property_generic` helper function (with no other helper functions) would be a better separation of concerns than keeping the logic inside of `set_property`.
</blockquote></details>

I can see why the distinction might not be considered worthwhile. For me, the two-fold distinction seemed worthy. The cases differ in two ways:

 - When the user references something that doesn't exist, this might reflect a typo or a case of out-of-order operations. There's not anything *intrinsically* bad about what they're doing. This deserves a warning.
 - When the user references something that simply cannot be done. I.e., they're going to try to access a property of something that exists already and cannot have a child property. That indicates a fundamental misunderstanding of the data/operation (or, the smaller possibility of a typo on one valid property resolves to a different, wrong valid property).

Finally, if I unify these cases, I'm forced to give them the same category. If I make them both warning, I really feel I'm underselling the fact that they're doing something that is simply wrong -- and I suspect people might be inclined to ignore warnings. On the other hand, if I promote them both to errors, then I dilute the error channel's contents.

(That said, I was able to reduce the logging *a bit*.)

So, does that sway you at all?

jwnimmer-tri (commented):
I am open to discussion on which of the code layouts will be superior to maintain over the long term -- that's basically a judgement call where reasonable people can differ.  I don't think the console severity has any bearing on that debate, though.  More detail inline, below.

> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-:-Nrm1Oh3EU5_QF9ZkY4Z:bkr4yut)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @SeanCurtis-TRI)

___
*[`src/index.js` line 715 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrkwAaiCWSQxGHvKhFY:-Nrm0LKR3DeFLORbc5EW:b-prj28e) ([raw file](https://github.com/meshcat-dev/meshcat/blob/3fb3b591bafa759b595587583978b59803a120f7/src/index.js#L715)):*
Every single console warning or error deserves attention from a developer who is trying to figure out why their code isn't working.  We should expect them to read all of them to make progress.  If there are so many that they need to start filtering things to find the wheat in the chaff, then we have much bigger problems.

I can buy the argument about "reporting real problems as warnings" is underselling.  That's why I suggested error for everything.

There's maybe something to be said that the two different colors (for different severities) will be easier to read somehow, but again -- there should be very few of these messages overall.  I don't buy that adding more control flow for that reason is a good trade-off.

> When the user references something that doesn't exist, this might reflect a typo or a case of out-of-order operations. There's not anything _intrinsically_ bad about what they're doing. This deserves a warning.

I don't buy that calling "set_property" and have it do nothing because the arguments were invalid can really be termed a "warning" in any real way.  They tried to set a property, but nothing happened.  Doing calls "out of order" is a mistake.

SeanCurtis-TRI:
*[`src/index.js` line 715 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrkwAaiCWSQxGHvKhFY:-Nrm3nvz1F5XI9BeUQYx:bb6yn4o) ([raw file](https://github.com/meshcat-dev/meshcat/blob/3fb3b591bafa759b595587583978b59803a120f7/src/index.js#L715)):*

Every single console warning or error deserves attention from a developer who is trying to figure out why their code isn't working.  We should expect them to read all of them to make progress.  If there are so many that they need to start filtering things to find the wheat in the chaff, then we have much bigger problems.

I can buy the argument about "reporting real problems as warnings" is underselling.  That's why I suggested error for everything.

There's maybe something to be said that the two different colors (for different severities) will be easier to read somehow, but again -- there should be very few of these messages overall.  I don't buy that adding more control flow for that reason is a good trade-off.

> When the user references something that doesn't exist, this might reflect a typo or a case of out-of-order operations. There's not anything _intrinsically_ bad about what they're doing. This deserves a warning.

I don't buy that calling "set_property" and have it do nothing because the arguments were invalid can really be termed a "warning" in any real way.  They tried to set a property, but nothing happened.  Doing calls "out of order" is a mistake.
</blockquote></details>

One last question: given the two very different reasons for failure; what error message would you print?

SeanCurtis-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[`src/index.js` line 715 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrkwAaiCWSQxGHvKhFY:-Nrp9WnO2ofpZ4TnNWle:bpnw6r0) ([raw file](https://github.com/meshcat-dev/meshcat/blob/3fb3b591bafa759b595587583978b59803a120f7/src/index.js#L715)):*

It's a little difficult to say for sure because the test cases aren't representative of the kinds of typos real users would make, but my intuition would go like the following:

(1) Reflect back what they were trying to do (i.e., the full path and property from the original message).  They will need to match up this error message with something they were doing with the C++ API.

(2) Include the full message from the JS exception.

(3) Maybe we could include how far we got before we crashed, in case the nesting is really deep (it is usually very deep?).

(4) Maybe we could include a dump of the current `target` so they can understand the list of valid properties (i.e., what they _should_ have said instead).

To validate those ideas, you could try to take rcory's current config file and play it against this new code.  If it has any typos, you could try to guess if the error would have been enough to help people who aren't you and I overcome those.
</blockquote></details>

By the way, I'm going to take my best stab at this, but I think your mental model may still misinform your expectations. From before:

```
  if (child in parent)
    return parent[child]
catch
  console error
  return null
```

This presumes the only condition for printing an error is an exception, and that is not true. The `else` clause of the if statement should also lead to an error as outlined above. I'm going to try my best to streamline it, but that's the crux of my concern. I don't think it lends itself well to squeezing because the two causes of error messages are both mechanically and semantically different.

That said, I may come up with something that does an amazing job squeezing things. We'll see.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @SeanCurtis-TRI)

___
*[`src/index.js` line 715 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrkwAaiCWSQxGHvKhFY:-NrpPzIN-n5c_Vip1C9m:bpr1mkg) ([raw file](https://github.com/meshcat-dev/meshcat/blob/3fb3b591bafa759b595587583978b59803a120f7/src/index.js#L715)):*

By the way, I'm going to take my best stab at this, but I think your mental model may still misinform your expectations. From before:

```
  if (child in parent)
    return parent[child]
catch
  console error
  return null
```

This presumes the only condition for printing an error is an exception, and that is not true. The `else` clause of the if statement should also lead to an error as outlined above. I'm going to try my best to streamline it, but that's the crux of my concern. I don't think it lends itself well to squeezing because the two causes of error messages are both mechanically and semantically different.

That said, I may come up with something that does an amazing job squeezing things. We'll see.
</blockquote></details>

Right, so it looks more like this:

```
exception_detail = ""
try
  if (child in parent)
    return parent[child]
catch
  exception_detail = ...
console.error(... + exception_detail)
return null
```

SeanCurtis-TRI (commented):
I've squeezed and co-located and refactored.

> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-:-Nrpg3mkAlaxBJ6nthOr:b-ol92r6)* status: 1 of 4 files reviewed, all discussions resolved (waiting on @jwnimmer-tri)

___
*[`src/index.js` line 715 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrkwAaiCWSQxGHvKhFY:-NrpTQAcAxPrE0ZBH6XK:b82jkjf) ([raw file](https://github.com/meshcat-dev/meshcat/blob/3fb3b591bafa759b595587583978b59803a120f7/src/index.js#L715)):*

Right, so it looks more like this:

```
exception_detail = ""
try
  if (child in parent)
    return parent[child]
catch
  exception_detail = ...
console.error(... + exception_detail)
return null
```
</blockquote></details>

Alrighty. Give this a spin.

jwnimmer-tri (commented):
I like it!  I will pull locally and re-run the test, but in the meantime let's get +@sammy-tri looped in for platform review, please.

> Reviewed 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-:-NrqJmXK-esUDPNhNUHV:bvj174o)* status: all files reviewed, 1 unresolved discussion (waiting on @jwnimmer-tri and @sammy-tri)

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrqJu8pBtnaftSRafaF:-NrqJu8pBtnaftSRafaG:b-uuz70t) (no related file):*
Working

Re-run some manual test(s).

___

jwnimmer-tri (approved):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177)* status: all files reviewed, 1 unresolved discussion (waiting on @sammy-tri and @SeanCurtis-TRI)

*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrqJu8pBtnaftSRafaF:-NrqMgLK2pL_jYAZILNW:b-c89tjf) (no related file):*

Working

Re-run some manual test(s).
</blockquote></details>

All good.  I ran Drake meshcat manual test as well.

___
___
*[`test/chained_properties.html` line 126 at r5](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrqKQTe2CDKr1OocdcY:-NrqKQTe2CDKr1OocdcZ:b-3phxnk) ([raw file](https://github.com/meshcat-dev/meshcat/blob/96ffa5a20fed3ed8fcbef486d2a77779537f27a1/test/chained_properties.html#L126)):*
> ```HTML
> 
>         console.info("\nSetting a property that doesn't exist at all. We should " +
>                      "get a warning indicating that and the guidance that the " +
> ```

nit Here and a few more times below, the advice says to expect a warning but we're all-errors now.  In an informal way this is fine because an error is a kind of warning, but I suspect you'd like to sharpen up these texts to say "error" instead of "warning" for crispness.

sammy-tri (changes requested):
> Reviewed 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177)* status: all files reviewed, 2 unresolved discussions (waiting on @SeanCurtis-TRI)

___
*[`Readme.md` line 263 at r5](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrqX9tR2wkm5C284Arp:-NrqX9tR2wkm5C284Arq:bo1ker) ([raw file](https://github.com/meshcat-dev/meshcat/blob/96ffa5a20fed3ed8fcbef486d2a77779537f27a1/Readme.md#L263)):*
> ```Markdown
>                         Properties not on the above list will be set directly on the <code>THREE.Object3D</code> object. This provides a powerful capability to customize the scene, but should be considered an advanced usage -- you're on your own to avoid any unwanted side-effects.
> 
>                         Properties can be *chained*. For example, for an object with a phong material (MeshPhongMaterial), we may want to tweak its shininess, making it duller. Shininess is not a property of the object itself, but the object's material. There is no *path* to that material, but the property name can include a property name chain, e.g., `material.shininess`. While setting the property, the chained properties will be evaluated in sequence, such that the final name in the chain is the property that receives the `value`.
> ```

This is rendering oddly for me in the browser window when I view the branch on github

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177)* status: all files reviewed, 2 unresolved discussions (waiting on @sammy-tri and @SeanCurtis-TRI)

___
*[`Readme.md` line 263 at r5](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrqX9tR2wkm5C284Arp:-Nrq_K5JDzo9NdFsG9VH:bnl0gre) ([raw file](https://github.com/meshcat-dev/meshcat/blob/96ffa5a20fed3ed8fcbef486d2a77779537f27a1/Readme.md#L263)):*
<details><summary><i>Previously, sammy-tri (Sam Creasey) wrote…</i></summary><blockquote>

This is rendering oddly for me in the browser window when I view the branch on github

</blockquote></details>

That's a really good point, I forgot to look at GitHub.

We probably need to use `<p>` breaks in lieu of blank lines.

SeanCurtis-TRI (commented):
All caught up.

> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-:-NrqhxBT5D-Lu56FRkTi:b-kk01nx)* status: all files reviewed, 1 unresolved discussion (waiting on @sammy-tri)

___
*[`Readme.md` line 263 at r5](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-NrqX9tR2wkm5C284Arp:-Nrqh9xL3n8WykSFNb-g:bb6un1j) ([raw file](https://github.com/meshcat-dev/meshcat/blob/96ffa5a20fed3ed8fcbef486d2a77779537f27a1/Readme.md#L263)):*

That's a really good point, I forgot to look at GitHub.

We probably need to use `<p>` breaks in lieu of blank lines.
</blockquote></details>

Done

sammy-tri (approved):
:lgtm:

> Reviewed 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/177#-:-Nrqj5DZCoKRaWNgzLc4:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SeanCurtis-TRI)

PR #187, opened 2025-02-03
Add InstancedMesh example

### LINKED ISSUE #141: InstancedMesh support
Hi, would it be possible to add InstancedMesh support with a per-instance `.instanceMatrix`? My use case is displaying a large number of ellipsoids (many more than `geometry.Ellipsoid` can handle).

I'm also happy to take a stab at this if there's a preferred way to implement.

siddancha:
Resolves #141 

The MeshCat API already implicitly supports threejs' [InstancedMesh](https://threejs.org/docs/#api/en/objects/InstancedMesh). Since InstancedMesh is a subclass of [Object3D](https://threejs.org/docs/#api/en/core/Object3D), the `set_object` API can be used to create an InstancedMesh of a given geometry and material.

This PR adds an example (in the `test` folder) that creates an InstancedMesh of BoxGeometries arranged in a 3x3x3 grid. Furthermore, the example updates the color buffer of the InstancedMesh using the `set_property` API. This showcases how one might update the colors (and similarly, poses) of instances in an InstancedMesh without deleting and creating a new InstancedMesh object.

After `set_object`:

After `set_property`:

### DIFF
--- a/test/instanced_mesh.html
+++ b/test/instanced_mesh.html
@@ -1,0 +1,167 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <meta charset=utf-8>
+    <title>MeshCat - InstancedMesh</title>
+</head>
+
+<body>
+    <div id="meshcat-pane">
+    </div>
+
+    <script src="../dist/main.min.js"></script>
+    <script>
+        var viewer = new MeshCat.Viewer(document.getElementById("meshcat-pane"));
+
+        // Set InstancedMesh object -- a grid of 3x3x3 BoxGeometries
+        viewer.handle_command({
+            type: "set_object",
+            path: "/meshcat/instanced_cubes",
+            object: {
+                metadata: { version: 4.5, type: "Object" },
+                geometries: [
+                    {
+                        uuid: "c2afd88b-ff9e-4408-9b94-b94e2079e2c8",
+                        type: "BoxGeometry",
+                        width: 0.2,
+                        height: 0.2,
+                        depth: 0.2
+                    }
+                ],
+                materials: [
+                    {
+                        uuid: "a63e3e66-c87a-4826-adee-2cf0fc04bf70", 
+                        type: "MeshPhongMaterial",
+                        color: 0xffffff,  // white
+                        side: 2,
+                        transparent: true,
+                        opacity: 0.5
+                    }
+                ],
+                object: {
+                    uuid: "d8497b0e-2529-4c5a-a477-00fc2c815349",
+                    type: "InstancedMesh",
+                    geometry: "c2afd88b-ff9e-4408-9b94-b94e2079e2c8",
+                    material: "a63e3e66-c87a-4826-adee-2cf0fc04bf70",
+                    instanceMatrix: {
+                        type: "InstancedBufferAttribute",
+                        array: [
+                            // Each line is a 4x4 transform matrix of a cube
+                            // The 4x4 matrices are flattened in column-major order
+
+                            // z = 0 layer
+                            1,0,0,0, 0,1,0,0, 0,0,1,0, 0,0,0,1,
+                            1,0,0,0, 0,1,0,0, 0,0,1,0, 0.3,0,0,1,
+                            1,0,0,0, 0,1,0,0, 0,0,1,0, 0.6,0,0,1,
+                            1,0,0,0, 0,1,0,0, 0,0,1,0, 0,0.3,0,1,
+                            1,0,0,0, 0,1,0,0, 0,0,1,0, 0.3,0.3,0,1,
+                            1,0,0,0, 0,1,0,0, 0,0,1,0, 0.6,0.3,0,1,
+                            1,0,0,0, 0,1,0,0, 0,0,1,0, 0,0.6,0,1,
+                            1,0,0,0, 0,1,0,0, 0,0,1,0, 0.3,0.6,0,1,
+                            1,0,0,0, 0,1,0,0, 0,0,1,0, 0.6,0.6,0,1,
+
+                            // z = 0.3 layer
+                            1,0,0,0, 0,1,0,0, 0,0,1,0, 0,0,0.3,1,
+                            1,0,0,0, 0,1,0,0, 0,0,1,0, 0.3,0,0.3,1,
+                            1,0,0,0, 0,1,0,0, 0,0,1,0, 0.6,0,0.3,1,
+                            1,0,0,0, 0,1,0,0, 0,0,1,0, 0,0.3,0.3,1,
+                            1,0,0,0, 0,1,0,0, 0,0,1,0, 0.3,0.3,0.3,1,
+                            1,0,0,0, 0,1,0,0, 0,0,1,0, 0.6,0.3,0.3,1,
+                            1,0,0,0, 0,1,0,0, 0,0,1,0, 0,0.6,0.3,1,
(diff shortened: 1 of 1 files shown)


joemasterjohn (commented):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/187)* status: 0 of 1 files reviewed, all discussions resolved

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/187#-OJU1abx0L0y0aRrW8V2:-OJU1abx0L0y0aRrW8V3:b4hiolu) (no related file):*
+@joemasterjohn for review.

joemasterjohn (changes requested):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/187)* status: 0 of 1 files reviewed, 3 unresolved discussions

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/187#-OJUEAZeEgL_ol7pBmjl:-OJUEAZeEgL_ol7pBmjm:bqtyimg) (no related file):*
Hi @siddancha, thanks for the contribution. Please consider some of the comments I left below.
___
*[`test/instanced_mesh.html` line 21 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/187#-OJU50ae36erS7i5mzcC:-OJU50ae36erS7i5mzcD:b-s9zrd5) ([raw file](https://github.com/meshcat-dev/meshcat/blob/bebe9599a74cd82edd71bf339305ee1f7b61912b/test/instanced_mesh.html#L21)):*
> ```HTML
>             type: "set_object",
>             path: "/meshcat/instanced_cubes",
>             object: {
> ```

I don't find much value in writing out JSON by hand for examples, especially JavaScript examples.
1) It has to be manually updated to keep up with newer THREE.js object format versioning.
2) It doesn't provide the user a method for generating their own JSON for their own purposes. It is still mysterious which properties must exist or how they should be formatted. Especially things like uuids etc.

I think an example written in JS that gives the user a way to spit out the JSON representation of whatever THREE.js object they're dealing with would be good. That would give them the reference to implement the object format correctly in whatever client code they are writing. Consider my update to your example here that does exactly that: https://github.com/joemasterjohn/meshcat/tree/instanced-mesh-example

If you agree, add me as a collaborator on your fork. I can push that commit (with co-author credits), and we can merge this very nice example. 
___
*[`test/instanced_mesh.html` line 100 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/187#-OJU2SIJ7KndhPzJN9MK:-OJU2SIJ7KndhPzJN9ML:bfbprj) ([raw file](https://github.com/meshcat-dev/meshcat/blob/bebe9599a74cd82edd71bf339305ee1f7b61912b/test/instanced_mesh.html#L100)):*
> ```HTML
>         // Set InstancedMesh properties.
>         // Here, we are updating the colors of the cubes.
>         viewer.handle_command({
> ```

This happens to work because the command executes before the first render call. Indeed if you put this call to `set_property` into a `setTimeout(..., 5000)` you won't see the color change. This is because any change to a buffer needs to be indicated by setting the `needsUpdate` attribute to true (see https://threejs.org/docs/#manual/en/introduction/How-to-update-things).

siddancha (commented):
Thanks for building on this example -- your interactive demo is awesome! I'm not convinced that the example should be written in native Javascript/THREE.js code vs. handwritten JSONs; more details in my reply to your comment.

> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/187#-:-OJV7dcM65jYtP03q8T9:b-xp4grc)* status: all files reviewed, 3 unresolved discussions (waiting on @joemasterjohn)

___
*[`test/instanced_mesh.html` line 21 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/187#-OJU50ae36erS7i5mzcC:-OJV0sIA5HsyVkK3bSHt:bcppa54) ([raw file](https://github.com/meshcat-dev/meshcat/blob/bebe9599a74cd82edd71bf339305ee1f7b61912b/test/instanced_mesh.html#L21)):*
<details><summary><i>Previously, joemasterjohn (Joe Masterjohn) wrote…</i></summary><blockquote>

I don't find much value in writing out JSON by hand for examples, especially JavaScript examples.
1) It has to be manually updated to keep up with newer THREE.js object format versioning.
2) It doesn't provide the user a method for generating their own JSON for their own purposes. It is still mysterious which properties must exist or how they should be formatted. Especially things like uuids etc.

I think an example written in JS that gives the user a way to spit out the JSON representation of whatever THREE.js object they're dealing with would be good. That would give them the reference to implement the object format correctly in whatever client code they are writing. Consider my update to your example here that does exactly that: https://github.com/joemasterjohn/meshcat/tree/instanced-mesh-example

If you agree, add me as a collaborator on your fork. I can push that commit (with co-author credits), and we can merge this very nice example.
</blockquote></details>

Hi @joemasterjohn! Thanks for building on this example -- your interactive demo is awesome! Correct me if I'm wrong, but I still think that writing out the JSON is what we need to demo this in the Meshcat API. Here's my reasoning:

1. I'm just following the other examples in the `test` folder, which also write out JSONs by hand. Instead, you created an `InstancedMesh` object in native Javascript/THREE.js, and are passing its JSON to the `set_object` command via the `.toJSON()` call. None of the other test examples use the `.toJSON()` call.
2. I think this makes sense because the way a non-JavaScript user would interact with the Meshcat API (e.g. via meshcat-python) is precisely by hand-writing JSONs and passing them to the `set_object` command. For example, I'm working on a PR to create InstancedMeshes in Meshcat from Drake: [SEE CODE HERE](https://github.com/siddancha/drake/blob/705cf98700ec29b5347c7ab90d4828756b9e4fe7/geometry/meshcat_types_internal.h#L351-L398). To implement this, I need to know what InstancedMesh's JSON looks like. Your native Javascript/THREE.js code unfortunately doesn't tell me much about what its JSON looks like and what fields I should set (THREE.js is to blame for that).

With this example (along the same lines as other examples) I'm hoping to expose what the [nominal format of Meshcat's API](https://github.com/meshcat-dev/meshcat?tab=readme-ov-file#api) would look like for InstancedMeshes. Does that make sense?
___
*[`test/instanced_mesh.html` line 100 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/187#-OJU2SIJ7KndhPzJN9MK:-OJV6t8r6L_5_h_MleMr:bz10ad3) ([raw file](https://github.com/meshcat-dev/meshcat/blob/bebe9599a74cd82edd71bf339305ee1f7b61912b/test/instanced_mesh.html#L100)):*
<details><summary><i>Previously, joemasterjohn (Joe Masterjohn) wrote…</i></summary><blockquote>

This happens to work because the command executes before the first render call. Indeed if you put this call to `set_property` into a `setTimeout(..., 5000)` you won't see the color change. This is because any change to a buffer needs to be indicated by setting the `needsUpdate` attribute to true (see https://threejs.org/docs/#manual/en/introduction/How-to-update-things).
</blockquote></details>

Agreed! I agree that we need to set `instanceMatrix.needsUpdate` after executing `set_property` commands, like you did in your branch. I will copy that once the previous comment is resolved.

siddancha (commented):
Maybe we could include both my hand-written JSON example and your THREE.js based example?

> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/187#-:-OJVEV11CN_WFUCTTBT-:b-bq0ltf)* status: all files reviewed, 3 unresolved discussions (waiting on @joemasterjohn)

PR #189, opened 2025-04-24
Upgrade three.js to v0.176.0

wrangelvid:
This change is [Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/189)


jwnimmer-tri (changes requested):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/189)* status: 0 of 5 files reviewed, 1 unresolved discussion

___
*[`dist/main.min.js.THIRD_PARTY_LICENSES.json` line 40 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/189#-OOcx3bI4PKg0Xjvsi0j:-OOcx3bI4PKg0Xjvsi0k:bffbog6) ([raw file](https://github.com/meshcat-dev/meshcat/blob/38dc6a1307084935796eab1b527559ee0bcd818f/dist/main.min.js.THIRD_PARTY_LICENSES.json#L40)):*
> ```JSON
>   {
>     "name": "three",
>     "version": "0.163.0",
> ```

Here we can see that the lock file has two different versions of three added to the bundle (176 and 163).  We have an invariant that there must be exactly one version for Meshcat.

This is usually because one of the other dependencies like `wwobjloader2` specifies a specific version of three, so usually we also need to upgrade those other dependencies in lockstep so that all bundled code agrees on a uniform version of three.

jwnimmer-tri (changes requested):
I confirmed that I could rebuild the minified dist file and it came out identical to what was pushed here.

> Reviewed 5 of 5 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/189#-:-OOhvsuj74-P43sGwhm-:bccioej)* status: all files reviewed, 1 unresolved discussion (waiting on @wrangelvid)

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/189#-OOhyt8K7ie_fCKOwgKx:-OOhyt8K7ie_fCKOwgKy:b-j2t1ro) (no related file):*
Working

All of the code and tests here look good.  Next step is to open a Drake PR with the git sha bump and do all of the Drake tests.  I'll open that PR momentarily.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/189)* status: all files reviewed, 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/189#-OOhyt8K7ie_fCKOwgKx:-OOi-n128MUMoL2Zolwx:b-3hstqa) (no related file):*
<details><summary><i>Previously, wrangelvid (David von Wrangel) wrote…</i></summary><blockquote>

Does it make sense to use the sha from #190, do the tests in drake s.t. we can land both in drake at once?
</blockquote></details>

Before any PR lands, all tests including Drake must pass, so the only different tactic would be to combine both PRs into one, but I think it's better to keep them separate.

jwnimmer-tri (approved):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/189)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @wrangelvid)

PR #193, opened 2025-10-04
Adds support for Line2 objects

ferrolho:
Adds support for importing and rendering Line2 objects, including LineMaterial and LineGeometry.  This allows for rendering lines with a configurable width.

Also adds a watch script to webpack for easier development.

**Figure 1.** Example of working code in MeshCat.jl (PR in that repo to follow).

Screenshot 2025-10-04 at 09 24 19


ferrolho:
Hi, @jwnimmer-tri! Can you have a look at this PR? Thanks! 🙏

SeanCurtis-TRI:
Quick update on documentaiton:

1. I skimmed the PR less quickly and realize all you've done is extend the parameter space for the `set_object` command. So, you can limit your self to that.
2. The test is definitely the more important detail -- it unblocks reviewers from being able to explicitly see the code being exercised.

PR #198, opened 2026-05-19
Provide texture "crawl" feature

SeanCurtis-TRI:
This is direct support of Drake's surface velocity implementation in which the surface of an object exhibits motion independent of the body's motion.

This implementation works by assigning novel properties to the appropriate materials (to provide uniform parameters to the shader). This required *deferred* evaluation of set_property(). For objects loaded asynchronously, a common sequence of of set_object() and set_property() calls would cause the property to be erased. Now, the call to set_property() is saved to be applied after the object is successfully loaded.

- - -
This change is [Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)


SeanCurtis-TRI (commented):
+a:@jwnimmer-tri for review, please.

> @SeanCurtis-TRI made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-:-Ot0vQUxEmNlptflYYD6:b-43r36z)* status: 0 of 9 files reviewed, all discussions resolved (waiting on jwnimmer-tri).

jwnimmer-tri (commented):
Per f2f this is basically on the right track.

However, I'm going to let the other two meshcat PRs merge first (embedded images, properties queue) and they cycle back here to review just the new protocol part by itself.

> @jwnimmer-tri made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-:-Otj1FO1-jTlvwZhK1P0:b-uy47mj)* status: 0 of 9 files reviewed, all discussions resolved.

SeanCurtis-TRI (commented):
Ok, we're up to date w.r.t. the preliminary PRs.

> @SeanCurtis-TRI made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-:-OuJzLIU-Ujw_utxrRPM:bhy9g9s)* status: 0 of 9 files reviewed, all discussions resolved (waiting on jwnimmer-tri).

jwnimmer-tri (changes requested):
> @jwnimmer-tri reviewed 1 file and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: 1 of 9 files reviewed, 2 unresolved discussions (waiting on SeanCurtis-TRI).

___
*[`Readme.md` line 337 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OublnhUBbp_1wVEW135:-OublnhVByNgrCQgDa7t:b7cann) ([raw file](https://github.com/meshcat-dev/meshcat/blob/27ad804427e7f9accddad21ce8dc41dd472f65fe/Readme.md#L337)):*
> ```Markdown
>                                 <p>The object(s) must have texture coordinates and texture(s) applied.</p>
>                                 <p>The "geometry frame" refers to the <strong>container</strong> frame as this typically contains the pose for the geometry as defined by the external application. For example, when setting <code>crawl_axis</code> on <code>/meshcat/textured_mesh</code>, the axis is expressed in the frame of <code>/meshcat/textured_mesh</code>. One *can* set the property to the renderable node as well
>                                 <code>/meshcat/textured_mesh/<object></code>, but the axis must still be expressed in the root geometry node's frame. We recommend treating these properties like other container-based properties like <code>position</code> or <code>visible</code>.</p>
> ```

See the preview at https://github.com/SeanCurtis-TRI/meshcat/blob/PR_crawling_texture/Readme.md.  Something about the style here goes awry; perhaps the `<object>` needs escaping.
___
*[`test/belt_bump.png` at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OuK-V3v2S82WJ-DSr9j:-OuK-V3v2S82WJ-DSr9k:b-3gxkki) ([raw file](https://github.com/meshcat-dev/meshcat/blob/27ad804427e7f9accddad21ce8dc41dd472f65fe/test/belt_bump.png)):*
For sanity within the test folder, at the very least we need to rename `belt_*` to `conveyor_belt_*` so that all of the related assets are "grouped" with a common prefix to their filename.  There's also an argument to be made that they should be named `crawling_texture*.*` to align with the test case that uses them.

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: 1 of 9 files reviewed, 2 unresolved discussions (waiting on jwnimmer-tri).

___
*[`Readme.md` line 337 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OublnhUBbp_1wVEW135:-OubqeWK4719mTBoG3ts:by5wxaj) ([raw file](https://github.com/meshcat-dev/meshcat/blob/27ad804427e7f9accddad21ce8dc41dd472f65fe/Readme.md#L337)):*

See the preview at https://github.com/SeanCurtis-TRI/meshcat/blob/PR_crawling_texture/Readme.md.  Something about the style here goes awry; perhaps the `<object>` needs escaping.
</blockquote></details>

Yep. &lt; and &gt; for the win.

I have to say, I hate the markdown-html mash up.
___
*[`test/belt_bump.png` at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OuK-V3v2S82WJ-DSr9j:-Oubs5ecDvp3MjVcmH2F:b-896fix) ([raw file](https://github.com/meshcat-dev/meshcat/blob/27ad804427e7f9accddad21ce8dc41dd472f65fe/test/belt_bump.png)):*

For sanity within the test folder, at the very least we need to rename `belt_*` to `conveyor_belt_*` so that all of the related assets are "grouped" with a common prefix to their filename.  There's also an argument to be made that they should be named `crawling_texture*.*` to align with the test case that uses them.
</blockquote></details>

Done.

jwnimmer-tri (changes requested):
> @jwnimmer-tri reviewed 2 files and all commit messages, made 4 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: 3 of 9 files reviewed, 5 unresolved discussions (waiting on SeanCurtis-TRI).

___
*[`Readme.md` line 336 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-Oubq5_81e5vO7TsNFqx:-Oubq5_81e5vO7TsNFqy:b2um43y) ([raw file](https://github.com/meshcat-dev/meshcat/blob/27ad804427e7f9accddad21ce8dc41dd472f65fe/Readme.md#L336)):*
> One *can* set the property to the renderable node as well `/meshcat/textured_mesh/<object>`, but the axis must still be expressed in the root geometry node's frame.

It seems to me that for our application needs we only care about this property on a "container" node.  Since setting it on a "renderable" node has this weird caveat about the frame being interpreted in a weird way, and requires special handling in the implementation, wouldn't it be simpler both in documentation complexity and implementation complexity to only allow setting it on container nodes?  Why do we go to lengths to support setting it on renderable nodes?
___
*[`Readme.md` line 325 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OuborAx8zqtR3cKSWVB:-OuborAx8zqtR3cKSWVC:b-o5owy) ([raw file](https://github.com/meshcat-dev/meshcat/blob/27ad804427e7f9accddad21ce8dc41dd472f65fe/Readme.md#L325)):*
> ```Markdown
>                             <li><code>bottom_color: number[3]</code> (only for the Background)
>                         </ul>
>                         <p>The following names are <strong>custom</strong> properties:</p>
> ```

nit The next thing after this sentence is not a custom property name.

(I suggest fixing this by actually putting the names front-and-center to start the section, instead of buried in the middle.)



_Code quote:_
```Markdown
The following names
```
___
*[`Readme.md` line 341 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OuboQyoEkE8sCFDxjTR:-OuboQyoEkE8sCFDxjTS:b-b69hxc) ([raw file](https://github.com/meshcat-dev/meshcat/blob/27ad804427e7f9accddad21ce8dc41dd472f65fe/Readme.md#L341)):*
> ```Markdown
>                             </li>
>                         </ul>
>                         <p>Any property name not in the list of convenience properties is evaluated as a literal property on the target object. Chained names (e.g., <code>material.shininess</code>), and are resolved left-to-right so the final name in the chain receives <code>value</code>.</p>
> ```

nit The language here needs an update for the new category.



_Suggestion:_
```Markdown
Any property name not in the list of convenience properties or custom properties is evaluated
```
___
*[`src/index.js` line 820 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OubrpTBEeA1m-g5oXak:-OubrpTBEeA1m-g5oXal:b-vpiysp) ([raw file](https://github.com/meshcat-dev/meshcat/blob/27ad804427e7f9accddad21ce8dc41dd472f65fe/src/index.js#L820)):*
> ```JavaScript
>     // object with respect to its parent container node. In all other cases,
>     // returns the Identity.
>     get_renderable_pose_in_container(target_path) {
> ```

It seems to me like if the target_path of a `crawl_axis` is ever *not* an `<object>`, then something has gone horribly wrong, because `forward_to_renderable` ensures that the target is a renderable node. Why are we guarding that failure condition with extra-complicated code and then silently ignoring it, instead of less code and/or tipping over immediately?

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI made 3 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: 3 of 9 files reviewed, 4 unresolved discussions (waiting on jwnimmer-tri).

___
*[`Readme.md` line 325 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OuborAx8zqtR3cKSWVB:-Oud1KN23u5ZqRbYHCU2:bmf15ls) ([raw file](https://github.com/meshcat-dev/meshcat/blob/27ad804427e7f9accddad21ce8dc41dd472f65fe/Readme.md#L325)):*
I pivoted a different direction based on the phrasing from the overview:

> Custom properties are Meshcat-specific properties that enable custom behavior and are not part of three.js.
___
*[`Readme.md` line 336 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-Oubq5_81e5vO7TsNFqx:-Oud1tyz092fWYMXaX0K:b-2riwje) ([raw file](https://github.com/meshcat-dev/meshcat/blob/27ad804427e7f9accddad21ce8dc41dd472f65fe/Readme.md#L336)):*

> One *can* set the property to the renderable node as well `/meshcat/textured_mesh/<object>`, but the axis must still be expressed in the root geometry node's frame.

It seems to me that for our application needs we only care about this property on a "container" node.  Since setting it on a "renderable" node has this weird caveat about the frame being interpreted in a weird way, and requires special handling in the implementation, wouldn't it be simpler both in documentation complexity and implementation complexity to only allow setting it on container nodes?  Why do we go to lengths to support setting it on renderable nodes?
</blockquote></details>

The primary reason is if anyone ever went looking for the property, they'd find it there. However, simply leaving that out of the documentation makes the most sense. Most people won't go looking for it.

If setting the property on the container is generally the right thing to do, then that's what the docs should say.
___
*[`src/index.js` line 820 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OubrpTBEeA1m-g5oXak:-Oud3XOOB5JiNrQcIROx:b-3e3xku) ([raw file](https://github.com/meshcat-dev/meshcat/blob/27ad804427e7f9accddad21ce8dc41dd472f65fe/src/index.js#L820)):*

It seems to me like if the target_path of a `crawl_axis` is ever *not* an `<object>`, then something has gone horribly wrong, because `forward_to_renderable` ensures that the target is a renderable node. Why are we guarding that failure condition with extra-complicated code and then silently ignoring it, instead of less code and/or tipping over immediately?
</blockquote></details>

Good point. Code evolved and this cul de sac got left behind.

jwnimmer-tri (commented):
LGTM on the design and approach to plumbing.  I haven't had time to go line-by-line through the code as a full feature review, though.

> @jwnimmer-tri reviewed 2 files and all commit messages, made 3 comments, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-:-OugZXCk9FS1eP3ukszD:b-kod6xv)* status: 3 of 9 files reviewed, 1 unresolved discussion (waiting on SeanCurtis-TRI).

___
*[`src/index.js` line 820 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OubrpTBEeA1m-g5oXak:-OugZ7Ey9BKzwxAhLSN9:bw8f965) ([raw file](https://github.com/meshcat-dev/meshcat/blob/27ad804427e7f9accddad21ce8dc41dd472f65fe/src/index.js#L820)):*

Good point. Code evolved and this cul de sac got left behind.
</blockquote></details>

This is fine, I guess.  I was hoping that the entire helper function would go away (and without any if-checks, for simplicity).  Having to jump around in the code between so many small helper functions makes the logic difficult to follow; having all of the logic inside the nice single-purpose helper functions might be a bit clearer.
___
*[`test/crawling_texture.html` line 269 at r5](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OugYnRf2ycCvOywHH0x:-OugYnRf2ycCvOywHH0y:bv6h5os) ([raw file](https://github.com/meshcat-dev/meshcat/blob/afbdb7ba77e858a82c543430f232b55f71e31e6b/test/crawling_texture.html#L269)):*
> ```HTML
>         });
>         load_comparison_box();
>         
> ```

nit Spurious whitespace?

SeanCurtis-TRI (commented):
+a:@joemasterjohn In Jeremy's absence, you are my go to guy to provide feature review for this kind of thing.  Thanks.

> @SeanCurtis-TRI made 2 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-:-Ouh_BWiEkiAyA6pruCL:b44p05e)* status: 3 of 9 files reviewed, all discussions resolved (waiting on joemasterjohn and jwnimmer-tri).

___
*[`src/index.js` line 820 at r3](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OubrpTBEeA1m-g5oXak:-Ouh_eb17qD0M-l2u2_9:b-pbdqz3) ([raw file](https://github.com/meshcat-dev/meshcat/blob/27ad804427e7f9accddad21ce8dc41dd472f65fe/src/index.js#L820)):*

This is fine, I guess.  I was hoping that the entire helper function would go away (and without any if-checks, for simplicity).  Having to jump around in the code between so many small helper functions makes the logic difficult to follow; having all of the logic inside the nice single-purpose helper functions might be a bit clearer.
</blockquote></details>

Agreed.

joemasterjohn (commented):
Pinging just to acknowledge I received this. I'll take a look, but I'm a bit bogged down with other Drake reviews, so I might be delayed.

> @joemasterjohn made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-:-OvBuPRb81-SR0aRgcO7:byxueey)* status: 2 of 9 files reviewed, all discussions resolved (waiting on jwnimmer-tri).

SeanCurtis-TRI:
No worries. I appreciate your being busy. Thanks for pinging.

jwnimmer-tri (changes requested):
> @jwnimmer-tri partially reviewed 7 files and all commit messages, and made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: all files reviewed, 6 unresolved discussions (waiting on joemasterjohn and SeanCurtis-TRI).

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjUqFX1CRaLffMPaRG:-OvjUqFX1CRaLffMPaRH:bigcug2) (no related file):*
Generally for changes aimed at Drake, we don't land meshcat PRs until downstream testing against an approved `meshcat_manual_test` has passed.  (I don't think we should land this until such a test is reviewed and passing.)
___
*[`src/index.js` line 985 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjWFrl7zaTgm5Q-hdK:-OvjWFrl7zaTgm5Q-hdL:b-uemi9v) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/src/index.js#L985)):*
> ```JavaScript
>         // We'll normalize to be safe with 32-bit floats.
>         const axis_N = axis_R.clone().applyMatrix3(R_NR).normalize();
>         const mats = Array.isArray(
> ```

BTW This line break smells weird (within the parens, instead of highlighting the ternary operator).
___
*[`src/index.js` line 1315 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjYmFr2nT9aqXSq33U:-OvjYmFr2nT9aqXSq33V:b1vj4dk) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/src/index.js#L1315)):*
> ```JavaScript
>         }
>     }
>     for (let child of node.children || []) {
> ```

It's not clear why we need a `|| []` fallback here. Nowhere else do we have a fallback for `children` being undefined. If I remove the `|| []`, the html test doesn't fail.
___
*[`src/index.js` line 2132 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjXttp7FLzSpUgP_W_:-OvjXttp7FLzSpUgP_Wa:b-om3y4a) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/src/index.js#L2132)):*
> ```JavaScript
>     // are explicitly hoisted to the renderable's container node. For all other
>     // *properties*, the original path is returned.
>     maybe_forward_to_renderable(path, property) {
> ```

I find this magic pretty confusing. I guess I still need to trace through the frame logic a big more, but for now it's a red flag that no other properties need this magic.  Or at least, why the magic isn't encapsulated within the `SceneNode.set_property` method -- why it leaks into the property dispatch.
___
*[`test/crawling_texture.html` line 12 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjbB3S--2F0xl-unUX:-OvjbB3S--2F0xl-unUY:bmqu597) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/test/crawling_texture.html#L12)):*
> ```HTML
>         You should see a conveyor belt glTF model and a plain box, both using
>         the same belt texture (the box applies the texture in the opposite
>         direction).</p>
> ```

nit The text afterwards assumes I know which box is the "box" and which box is the glTF, but we never tell the reader how to know which one is which. We should explain.
___
*[`test/crawling_texture.html` line 32 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjbtpzEV6Dq-HLNFbs:-OvjbtpzEV6Dq-HLNFbt:bn2je7c) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/test/crawling_texture.html#L32)):*
> ```HTML
>         the same speed (in the world frame). If they are moving at different
>         speeds, it means an important contributor to the mapping of texture
>         coordinates to world frame has been missed.
> ```

For me, the top (and bottom) of the boxes are moving at the same speed, but the thin ends are not.  The longer box looks correct, but the shorter box texture doesn't look correct at all.

SeanCurtis-TRI (commented):
-a:@joemasterjohn 

Thanks for the review, Jeremy. PTAL.

> @SeanCurtis-TRI made 5 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-:-OvlH3m90-56ixQ-PkPG:bhyvxcs)* status: all files reviewed, 4 unresolved discussions (waiting on jwnimmer-tri).

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjUqFX1CRaLffMPaRG:-OvlHBmB69TVE5YYm4G_:bj3knzd) (no related file):*

Generally for changes aimed at Drake, we don't land meshcat PRs until downstream testing against an approved `meshcat_manual_test` has passed.  (I don't think we should land this until such a test is reviewed and passing.)
</blockquote></details>

Ok, I'll progress with the other dynamics portions of the conveyor belt and ping against this when we get to visualization.

Nevertheless, we should get this is in a fully reviewed state prior to that.
___
*[`src/index.js` line 1315 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjYmFr2nT9aqXSq33U:-OvlCDy092hMlrMM69g0:byt1p2e) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/src/index.js#L1315)):*

It's not clear why we need a `|| []` fallback here. Nowhere else do we have a fallback for `children` being undefined. If I remove the `|| []`, the html test doesn't fail.
</blockquote></details>

Agreed -- the `children` property is never `null` or `undefined`.
___
*[`src/index.js` line 2132 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjXttp7FLzSpUgP_W_:-OvkiqwWCCehVDZ7KZL-:bsm73fy) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/src/index.js#L2132)):*

I find this magic pretty confusing. I guess I still need to trace through the frame logic a big more, but for now it's a red flag that no other properties need this magic.  Or at least, why the magic isn't encapsulated within the `SceneNode.set_property` method -- why it leaks into the property dispatch.
</blockquote></details>

Ok, I've buried it in the set node.

The virtue of having it here, is that it guarantees the renderable node exists before entering SceneNode code and then we only enter the renderable node.

Take a look at the localization.
___
*[`test/crawling_texture.html` line 32 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjbtpzEV6Dq-HLNFbs:-OvlDuFSDDYIl6Bn95ua:b-vnk6bq) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/test/crawling_texture.html#L32)):*

For me, the top (and bottom) of the boxes are moving at the same speed, but the thin ends are not.  The longer box looks correct, but the shorter box texture doesn't look correct at all.


</blockquote></details>

This is actually an optical illusion.

The problem is the spatial density of the texture is higher than elsewhere. You see *information* going by at a higher rate, but that's only because the texture is spatially compressed. THe consequence, however, is that your brain interprets that as a faster surface.

The counter evidence is to watch the arrow traverse this face from side to side. Compare that to the glTF model on its right (in this view). It also has an arrow that moves along its edge. 

If you watch you'll see the period at which an arrow crosses this apparently problematic leading face matches the period when a new arrow appears on the adjacent face. Those two arrows stay in perfect lock step.

I fell victim to the same optical illusion. What helped me resolve it was the `conveyor_belt` demo in the other test branch. One box uses the texture as a tank tread. As the box "slides" along, the texture gave me the impression that it was moving at a higher speed. However, when I was able to look at the box's bottom, I could see that it's not true. In that case, I had a fixed reference point (the ground grid). I was able to see that even with the box moving across the grid, the features of the texture remained fixed relative to the ground grid. The texture was moving at exactly the same speed as the box, it was just the relative scale of the textures that deceived me.

jwnimmer-tri (changes requested):
> @jwnimmer-tri reviewed 1 file and all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: 7 of 9 files reviewed, 4 unresolved discussions (waiting on SeanCurtis-TRI).

___
*[`test/crawling_texture.html` line 12 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjbB3S--2F0xl-unUX:-Ovlf7NpA3Tyngr4FTTa:bj8zczn) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/test/crawling_texture.html#L12)):*

nit The text afterwards assumes I know which box is the "box" and which box is the glTF, but we never tell the reader how to know which one is which. We should explain.
</blockquote></details>

I don't see any "green-ish" in the demo.  A definitive way of explaining which one is which would be saying "longer box" vs "shorter box" in the docs here, or even better placing a 2d text on-screen next to them ala texture_text.html.
___
*[`test/crawling_texture.html` line 32 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjbtpzEV6Dq-HLNFbs:-OvlfLO_7f17xeW9ftR4:bncet94) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/test/crawling_texture.html#L32)):*

This is actually an optical illusion.

The problem is the spatial density of the texture is higher than elsewhere. You see *information* going by at a higher rate, but that's only because the texture is spatially compressed. THe consequence, however, is that your brain interprets that as a faster surface.

The counter evidence is to watch the arrow traverse this face from side to side. Compare that to the glTF model on its right (in this view). It also has an arrow that moves along its edge. 

If you watch you'll see the period at which an arrow crosses this apparently problematic leading face matches the period when a new arrow appears on the adjacent face. Those two arrows stay in perfect lock step.

I fell victim to the same optical illusion. What helped me resolve it was the `conveyor_belt` demo in the other test branch. One box uses the texture as a tank tread. As the box "slides" along, the texture gave me the impression that it was moving at a higher speed. However, when I was able to look at the box's bottom, I could see that it's not true. In that case, I had a fixed reference point (the ground grid). I was able to see that even with the box moving across the grid, the features of the texture remained fixed relative to the ground grid. The texture was moving at exactly the same speed as the box, it was just the relative scale of the textures that deceived me.
</blockquote></details>

I wasn't talking about if I change the crawl axis.  With the default axis, the thin side of one of the two boxes doesn't look correct at all.  The texture is just garbage, not even the stripes.

jwnimmer-tri (changes requested):
> @jwnimmer-tri made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: 7 of 9 files reviewed, 4 unresolved discussions (waiting on SeanCurtis-TRI).

___
*[`test/crawling_texture.html` line 32 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjbtpzEV6Dq-HLNFbs:-OvlgPLj2IVhOvQLJQDG:b-z36rwc) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/test/crawling_texture.html#L32)):*

I wasn't talking about if I change the crawl axis.  With the default axis, the thin side of one of the two boxes doesn't look correct at all.  The texture is just garbage, not even the stripes.
</blockquote></details>

In other words, I would have expected the thin side of the primitive box vs the glTF box to look roughly similar, modulo coloration and velocity.  But they look nothing alike.

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: 7 of 9 files reviewed, 4 unresolved discussions (waiting on jwnimmer-tri).

___
*[`test/crawling_texture.html` line 12 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjbB3S--2F0xl-unUX:-OvoWGua8vZsMbl2JPG9:b3v0vhs) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/test/crawling_texture.html#L12)):*

I don't see any "green-ish" in the demo.  A definitive way of explaining which one is which would be saying "longer box" vs "shorter box" in the docs here, or even better placing a 2d text on-screen next to them ala texture_text.html.
</blockquote></details>

This is what I clearly see:



I'm not sure what you see such that the box isn't "greenish". I stayed away from longer/shorter because I wanted the user to be able to distinguish the two boxes, even when they can't see the ends.
___
*[`test/crawling_texture.html` line 32 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjbtpzEV6Dq-HLNFbs:-OvoWQmx4NoHejW7k-BY:b8wuzrf) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/test/crawling_texture.html#L32)):*

In other words, I would have expected the thin side of the primitive box vs the glTF box to look roughly similar, modulo coloration and velocity.  But they look nothing alike.
</blockquote></details>

Your expectations mislead you.

If you set the crawl speed to zero, I believe that your expectations for how the box is textured won't match your expectations. That disconnect has nothing to do with this PR and has everything to do with how meshcat assigns texture coordinates to the box.



In this image, I've disconnected the crawl axis from the reference box entirely. It plays no role. You'll note that the texture is applied to each face in its entirety, regardless of the actual size/shape of the box face. These are the default texture coordinates of the box. This is the effect you get if you simply apply a moving texture to a primitive box. There is nothing that the code in this PR can or should do to alter the fundamental texture coordinate layout. What this PR does do *with* those texture coordinates is correct.

What you like so much about the conveyor belt is the result of hand-crafting the texture coordinates so that a pixel from the image has the same measure in the geometry frame on all of the faces (well, that's a bit of a lie, but true enough).

Now, if I'm still missing your point, this would most easily be handled with a f2f where you show me your screen an point out what it is that's bothering you. As I look at it, I see nothing anomalous.

I suspect the true problem is communicating to users the correct expectation. I'm going to twiddle with the test as is and see if an alternate representation might still serve the desired testing while avoiding this kind of confusion.

The point that should *ultimately* be emphasized is that crawling textures will *generally* not look correct. The only way it will have the *intended* effect is if the texture coordinates are applied to your geometry in a meaningful way. (Again, this is the primary reason why I wasn't originally a fan of this approach and favored something more abstract and glyph like that would work regardless of texture layout.) The only reason I'm sticking with this is that when things *are* configured well, the visual effect is *especially* compelling. Kind of a double-or-nothing affair.

jwnimmer-tri (changes requested):
> @jwnimmer-tri made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: 7 of 9 files reviewed, 4 unresolved discussions (waiting on SeanCurtis-TRI).

___
*[`test/crawling_texture.html` line 12 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjbB3S--2F0xl-unUX:-Ovoh_CO6I5bb-vPjuqT:b-y3deu) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/test/crawling_texture.html#L12)):*

This is what I clearly see:



I'm not sure what you see such that the box isn't "greenish". I stayed away from longer/shorter because I wanted the user to be able to distinguish the two boxes, even when they can't see the ends.
</blockquote></details>

Re: green -- in a different thread I posted a screenshot, where you can see my color is different.
___
*[`test/crawling_texture.html` line 32 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjbtpzEV6Dq-HLNFbs:-OvogzJl2g_FjrefUgXf:b-lqo1l4) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/test/crawling_texture.html#L32)):*

Your expectations mislead you.

If you set the crawl speed to zero, I believe that your expectations for how the box is textured won't match your expectations. That disconnect has nothing to do with this PR and has everything to do with how meshcat assigns texture coordinates to the box.



In this image, I've disconnected the crawl axis from the reference box entirely. It plays no role. You'll note that the texture is applied to each face in its entirety, regardless of the actual size/shape of the box face. These are the default texture coordinates of the box. This is the effect you get if you simply apply a moving texture to a primitive box. There is nothing that the code in this PR can or should do to alter the fundamental texture coordinate layout. What this PR does do *with* those texture coordinates is correct.

What you like so much about the conveyor belt is the result of hand-crafting the texture coordinates so that a pixel from the image has the same measure in the geometry frame on all of the faces (well, that's a bit of a lie, but true enough).

Now, if I'm still missing your point, this would most easily be handled with a f2f where you show me your screen an point out what it is that's bothering you. As I look at it, I see nothing anomalous.

I suspect the true problem is communicating to users the correct expectation. I'm going to twiddle with the test as is and see if an alternate representation might still serve the desired testing while avoiding this kind of confusion.

The point that should *ultimately* be emphasized is that crawling textures will *generally* not look correct. The only way it will have the *intended* effect is if the texture coordinates are applied to your geometry in a meaningful way. (Again, this is the primary reason why I wasn't originally a fan of this approach and favored something more abstract and glyph like that would work regardless of texture layout.) The only reason I'm sticking with this is that when things *are* configured well, the visual effect is *especially* compelling. Kind of a double-or-nothing affair.
</blockquote></details>

As far as the visual confusion, another solution would be to tell people running this test to only look at the top surface, and to ignore the sides.

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: 7 of 9 files reviewed, 4 unresolved discussions (waiting on jwnimmer-tri).

___
*[`test/crawling_texture.html` line 12 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjbB3S--2F0xl-unUX:-OvomBfJ7c8ntxO2FD1K:b-q37k36) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/test/crawling_texture.html#L12)):*

Re: green -- in a different thread I posted a screenshot, where you can see my color is different.
</blockquote></details>

Yes...I turned it green at the same time I documented it as being greenish. If I understand you, the image you posted predates the text and the color shift. In this revision, you can scroll down and see the explicit declaration where I've set the "color" property on the reference box.
___
*[`test/crawling_texture.html` line 32 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjbtpzEV6Dq-HLNFbs:-OvomX-E3HBFh_wi290g:bhojm0q) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/test/crawling_texture.html#L32)):*

As far as the visual confusion, another solution would be to tell people running this test to only look at the top surface, and to ignore the sides.
</blockquote></details>

Well, that somewhat undercuts the point of the test. Part of the test is to change the axis and see that it has the same effect on each of the shapes. Saying, "ignore the sides" undercuts that purpose. Again, I'm tinkering with an alternative.

jwnimmer-tri (commented):
> @jwnimmer-tri made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: 7 of 9 files reviewed, 3 unresolved discussions (waiting on SeanCurtis-TRI).

___
*[`test/crawling_texture.html` line 12 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjbB3S--2F0xl-unUX:-Ovon2z5AiLP9J2W4srK:b15qd2o) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/test/crawling_texture.html#L12)):*

Yes...I turned it green at the same time I documented it as being greenish. If I understand you, the image you posted predates the text and the color shift. In this revision, you can scroll down and see the explicit declaration where I've set the "color" property on the reference box.
</blockquote></details>

Ah, okay.  I didn't pull the latest.

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: 6 of 9 files reviewed, 3 unresolved discussions (waiting on jwnimmer-tri).

___
*[`test/crawling_texture.html` line 12 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjbB3S--2F0xl-unUX:-OvpdqGi-XOxP3pT8jfB:bpj95e7) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/test/crawling_texture.html#L12)):*

Ah, okay.  I didn't pull the latest.
</blockquote></details>

I've overhauled the test page to help clarify and head off confusion at the pass. In spirit, it's the same thing, in practice it hopefully sticks the landing better. PTAL.

jwnimmer-tri (changes requested):
> @jwnimmer-tri partially reviewed 2 files and all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: 8 of 9 files reviewed, 2 unresolved discussions (waiting on SeanCurtis-TRI).

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjUqFX1CRaLffMPaRG:-Ovtqanu7fzU7PH6kugT:b3dp5rd) (no related file):*

Ok, I'll progress with the other dynamics portions of the conveyor belt and ping against this when we get to visualization.

Nevertheless, we should get this is in a fully reviewed state prior to that.
</blockquote></details>

Can you set to "working" to clear the red bubble from me?

From my point of view, we only need the meshcat_manual_test coverage before we land this.
___
*[`src/index.js` line 2132 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjXttp7FLzSpUgP_W_:-OvtrKtt-2U9FNFX0O_T:b-e525i7) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/src/index.js#L2132)):*

Ok, I've buried it in the set node.

The virtue of having it here, is that it guarantees the renderable node exists before entering SceneNode code and then we only enter the renderable node.

Take a look at the localization.
</blockquote></details>

WDYT about https://github.com/SeanCurtis-TRI/meshcat/pull/2?

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: 8 of 9 files reviewed, 2 unresolved discussions (waiting on jwnimmer-tri and SeanCurtis-TRI).

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjUqFX1CRaLffMPaRG:-Ovu3pDF4AiV2FnszReg:b-w7gpc0) (no related file):*

Can you set to "working" to clear the red bubble from me?

From my point of view, we only need the meshcat_manual_test coverage before we land this.
</blockquote></details>

For cross reference: https://github.com/RobotLocomotion/drake/pull/24665
___
*[`src/index.js` line 2132 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjXttp7FLzSpUgP_W_:-Ovu3rCc-xZhH87jp45J:b-o6tqx1) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f2b5c0c5921e12c556217ee5f4ae219cdfd7ca45/src/index.js#L2132)):*

WDYT about https://github.com/SeanCurtis-TRI/meshcat/pull/2?
</blockquote></details>

Very nice. That's what I ultimately wanted, but I hadn't started in the right direction and ended up painting myself into a corner.

jwnimmer-tri (approved):
:lgtm: feature.

> @jwnimmer-tri reviewed 1 file and all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-:-Ovu4M4G32EuM1tAI7A1:b-n37teh)* status: all files reviewed, 2 unresolved discussions (waiting on SeanCurtis-TRI).

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-Ovu4Mor96RpSn9d1_c0:-Ovu4Mor96RpSn9d1_c1:b-d71060) (no related file):*
nit Needs a `rebuild.sh`.

jwnimmer-tri (commented):
> @jwnimmer-tri reviewed 1 file and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: all files reviewed, 1 unresolved discussion (waiting on SeanCurtis-TRI).

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: all files reviewed, 1 unresolved discussion (waiting on SeanCurtis-TRI).

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjUqFX1CRaLffMPaRG:-OvuLsEZ0rledRP74Rhl:b-pdv853) (no related file):*

For cross reference: https://github.com/RobotLocomotion/drake/pull/24665
</blockquote></details>

@jwnimmer-tri is that meshcat PR sufficient proof of life to merge here?

jwnimmer-tri (approved):
> @jwnimmer-tri made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: all files reviewed, 1 unresolved discussion (waiting on SeanCurtis-TRI).

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvjUqFX1CRaLffMPaRG:-OvuME-HDHRT9urv3XyY:b-njcn09) (no related file):*

@jwnimmer-tri is that meshcat PR sufficient proof of life to merge here?
</blockquote></details>

Yes, but this still need platform review of this.

SeanCurtis-TRI:
+a:@xuchen-han for platform review, please.

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on xuchen-han).

SeanCurtis-TRI:
Blocking for platform review -- Reviewable doesn't recognize that as a requirement.

xuchen-han (approved):
:lgtm: with the caveat that I'm a javascript noob.

> @xuchen-han partially reviewed 9 files and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-:-Ovv_EGfDO6OC3LGmCye:b-93js2e)* status: all files reviewed, 4 unresolved discussions (waiting on SeanCurtis-TRI).

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-Ovv_APR77PcmkfsjsU5:-Ovv_APR77PcmkfsjsU6:bp1qchp) (no related file):*

I don't see the reference boxes in the test. What am I missing?
___
*[`src/index.js` line 991 at r10](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvvYSil2xxd-5QL4pdg:-OvvYSil2xxd-5QL4pdh:b-gjw4w0) ([raw file](https://github.com/meshcat-dev/meshcat/blob/57f6781a2d8f8737be36716e6aaa061ee159a7c4/src/index.js#L991)):*
> ```JavaScript
>         const R_NR = R_RN.clone().transpose();
>         // We'll normalize to be safe with 32-bit floats.
>         const axis_N = axis_R.clone().applyMatrix3(R_NR).normalize();
> ```

nit, do we need a zero guard here?
___
*[`src/index.js` line 996 at r10](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvvbZ5WAS_uIewGo5v1:-OvvbZ5WAS_uIewGo5v2:bpt6r4f) ([raw file](https://github.com/meshcat-dev/meshcat/blob/57f6781a2d8f8737be36716e6aaa061ee159a7c4/src/index.js#L996)):*
> ```JavaScript
>             : [node.material];
>         for (const mat of mats) {
>             mat.meshcat_crawl_axis = axis_N.clone();
> ```

BTW am I correct in understanding that if two geometries share the same material then we can't control the crawling independently? i.e. I can't have one conveyor crawl and the other stay still.

jwnimmer-tri (commented):
> @jwnimmer-tri made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: all files reviewed, 4 unresolved discussions (waiting on SeanCurtis-TRI and xuchen-han).

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-Ovv_APR77PcmkfsjsU5:-Ovvd04lC6DwqbQVEtYF:b-3vdr1c) (no related file):*
<details><summary><i>Previously, xuchen-han (Xuchen Han) wrote…</i></summary><blockquote>


I don't see the reference boxes in the test. What am I missing?
</blockquote></details>

Did you perhaps open the page as a `file://` URL instead of `http://`?  There is a `test/Readme.md` that says how to run a local http server, so that assets work.

xuchen-han (approved):
> @xuchen-han made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: all files reviewed, 3 unresolved discussions (waiting on SeanCurtis-TRI).

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-Ovv_APR77PcmkfsjsU5:-Ovvn82p2PlSb07OBuKA:b1f2ij0) (no related file):*

Did you perhaps open the page as a `file://` URL instead of `http://`?  There is a `test/Readme.md` that says how to run a local http server, so that assets work.
</blockquote></details>

Yes that's it. Thanks!

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI made 2 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: all files reviewed, 1 unresolved discussion (waiting on xuchen-han).

___
*[`src/index.js` line 991 at r10](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvvYSil2xxd-5QL4pdg:-OvzBs2k5Q87T78X1GDV:bh9lg0j) ([raw file](https://github.com/meshcat-dev/meshcat/blob/57f6781a2d8f8737be36716e6aaa061ee159a7c4/src/index.js#L991)):*
<details><summary><i>Previously, xuchen-han (Xuchen Han) wrote…</i></summary><blockquote>

nit, do we need a zero guard here?
</blockquote></details>

We don't need to *guard* per se. We won't leak NaNs in. What we'll get instead is a zero vector.

It's true that a vector of epsilon magnitude will have its actual direction dominated by round-off when it gets normalized. But every value will produce *a* vector. 

I've gone ahead and added a note to the documentation to make that explicit.
___
*[`src/index.js` line 996 at r10](https://reviewable.io/reviews/meshcat-dev/meshcat/198#-OvvbZ5WAS_uIewGo5v1:-OvzBIgF2ggbmi0i2UFc:b-69fd6q) ([raw file](https://github.com/meshcat-dev/meshcat/blob/57f6781a2d8f8737be36716e6aaa061ee159a7c4/src/index.js#L996)):*
<details><summary><i>Previously, xuchen-han (Xuchen Han) wrote…</i></summary><blockquote>

BTW am I correct in understanding that if two geometries share the same material then we can't control the crawling independently? i.e. I can't have one conveyor crawl and the other stay still.
</blockquote></details>

That is *strictly* true, but not practically relevant.

1) Generally, separate objects instantiate independent materials. Even if the specified material is identical.
2) In the case where multiple meshes specified within a single file (e.g., glTF) use the same Material, they *will* share the same material. But we are intentionally applying the crawl to *all* meshes contained in the glTF. So, we're treating them the same by design.

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: 8 of 9 files reviewed, all discussions resolved (waiting on jwnimmer-tri and xuchen-han).

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on SeanCurtis-TRI).

jwnimmer-tri (approved):
> @jwnimmer-tri reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/198)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on SeanCurtis-TRI).

PR #200, opened 2026-05-28
Defer property replay until objects materialize

SeanCurtis-TRI:
Call to set_property() has been made more robust to handle objects that load asynchronously. It is still necessary to call set_object() *before* set_property(), but the effective act is the same as if all set_object()
invocations were effectively synchronized with invocations of set_property().

This also includes careful coordination with the property chaining. The key is to distinguish when a property should be queued and when it should not be.

The tests for chained_properties.html has been updated to support a similar test pass/fail feedback in browser as deferred_property_replay.html. This reduces the cognitive demand on developers to determine whether the test was successful in its assertions or not.

A minor refinement of set_object(). set_object("/foo/bar",...) and set_object("/foo/bar/<object>", ...) are now equivalent.

- - -
This change is [Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200)


SeanCurtis-TRI:
+a:@jwnimmer-tri for feature review, please.

jwnimmer-tri (changes requested):
> @jwnimmer-tri reviewed 1 file and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200)* status: 1 of 3 files reviewed, 4 unresolved discussions (waiting on SeanCurtis-TRI).

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtjTzjN77Y5k-R_dVjH:-OtjTzjN77Y5k-R_dVjI:buyhnmp) (no related file):*
The `Readme.md` says this, which is out of date now:

> If a name in the property chain is missing, an error message will be printed to the console and no value will be assigned.

Similarly the `chained_properties.html` test case has this, which is out of date now:

> ```js
>         console.info("\nSetting a property that doesn't exist at all. We should " +
>                      "get an error indicating that fact and the guidance that the " +
>                      "property has not been set.");
> ```
___
*[`src/index.js` line 777 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtjS6ej9qnqoIOf9vVK:-OtjS6ej9qnqoIOf9vVL:b-1lwl8s) ([raw file](https://github.com/meshcat-dev/meshcat/blob/19b509df6c231b5e22598375ab08a248267cf746/src/index.js#L777)):*
> ```JavaScript
>         // Store property for replay when object materializes (e.g., after async
>         // glTF load)
>         this.pending_properties[property] = {value, target_path};
> ```

It seems like we add the property to "pending" even if it was correctly set. Why is that? The purpose of `pending` seems to be to handle the case of objects not existing yet, we don't seem to be checking for that.

This seems to be doing the `set_property` operation twice, which both wastes time and makes it difficult to reason about sequencing in case the same property is being repeatedly changed.
___
*[`src/index.js` line 823 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtjT0_YDNNYsv15EwdU:-OtjT0_YDNNYsv15EwdV:byg20ml) ([raw file](https://github.com/meshcat-dev/meshcat/blob/19b509df6c231b5e22598375ab08a248267cf746/src/index.js#L823)):*
> ```JavaScript
>             // The child doesn't exist yet; we'll queue the property
>             // assignment to handle asynchronous creation.
>             this.pending_properties[property] = {value, target_path};
> ```

The `property` we are storing here has already been canonicalized (replacing `[]` with `.` and removing any leading `.`). However, up above during `set_property` we store the _non-canonical_ property as the dictionary key.

In case canonicalization actually changed the string, that means there will be two different property values stored, with different-but-equivalent keys. If the user sets the same property via multiple different aliases that all canonicalize to the same name, then the "last one wins" invariant might get broken.
___
*[`src/index.js` line 860 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtjUQek2_XQdxoxgv2k:-OtjUQek2_XQdxoxgv2l:bxstlqb) ([raw file](https://github.com/meshcat-dev/meshcat/blob/19b509df6c231b5e22598375ab08a248267cf746/src/index.js#L860)):*
> ```JavaScript
>         for (let property in queued_props) {
>             let {value, target_path} = queued_props[property];
>             this.set_property(property, value, target_path);
> ```

If the property was _still_ unknown, I believe `this.set_property` will add it back into the queue, indefinitely.  Shouldn't the bad `set_property` eventually time out with an error, so that (1) the user gets feedback and (2) we don't leak memory (and time spent retrying) indefinitely?

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200)* status: 1 of 3 files reviewed, 4 unresolved discussions (waiting on jwnimmer-tri).

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtjTzjN77Y5k-R_dVjH:-OtjZYlU5mBfbUBnXQax:b-e41jbr) (no related file):*

The `Readme.md` says this, which is out of date now:

> If a name in the property chain is missing, an error message will be printed to the console and no value will be assigned.

Similarly the `chained_properties.html` test case has this, which is out of date now:

> ```js
>         console.info("\nSetting a property that doesn't exist at all. We should " +
>                      "get an error indicating that fact and the guidance that the " +
>                      "property has not been set.");
> ```
</blockquote></details>

Well spotted. I actually overhauled the relationship between property chains and property replay in order to keep the error conditions clean.
___
*[`src/index.js` line 777 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtjS6ej9qnqoIOf9vVK:-Otjnk8ZAyxIHDPHObkX:bvv5z3a) ([raw file](https://github.com/meshcat-dev/meshcat/blob/19b509df6c231b5e22598375ab08a248267cf746/src/index.js#L777)):*

It seems like we add the property to "pending" even if it was correctly set. Why is that? The purpose of `pending` seems to be to handle the case of objects not existing yet, we don't seem to be checking for that.

This seems to be doing the `set_property` operation twice, which both wastes time and makes it difficult to reason about sequencing in case the same property is being repeatedly changed.
</blockquote></details>

As part of the effort to better coordinate with chained properties, this has been clarified, I hope.
___
*[`src/index.js` line 823 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtjT0_YDNNYsv15EwdU:-OtjxmwH7OxdhKdIIkOB:b3jewee) ([raw file](https://github.com/meshcat-dev/meshcat/blob/19b509df6c231b5e22598375ab08a248267cf746/src/index.js#L823)):*

The `property` we are storing here has already been canonicalized (replacing `[]` with `.` and removing any leading `.`). However, up above during `set_property` we store the _non-canonical_ property as the dictionary key.

In case canonicalization actually changed the string, that means there will be two different property values stored, with different-but-equivalent keys. If the user sets the same property via multiple different aliases that all canonicalize to the same name, then the "last one wins" invariant might get broken.
</blockquote></details>

This has now changed. PTAL.
___
*[`src/index.js` line 860 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtjUQek2_XQdxoxgv2k:-OtjzNE30njrTuFvZ20Q:bee9dtr) ([raw file](https://github.com/meshcat-dev/meshcat/blob/19b509df6c231b5e22598375ab08a248267cf746/src/index.js#L860)):*

If the property was _still_ unknown, I believe `this.set_property` will add it back into the queue, indefinitely.  Shouldn't the bad `set_property` eventually time out with an error, so that (1) the user gets feedback and (2) we don't leak memory (and time spent retrying) indefinitely?
</blockquote></details>

I believe that we're ok in the new infrastructure. PTAL.

jwnimmer-tri (changes requested):
> @jwnimmer-tri reviewed 1 file, made 4 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200)* status: 1 of 5 files reviewed, 4 unresolved discussions (waiting on SeanCurtis-TRI).

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtjTzjN77Y5k-R_dVjH:-OtkPBxd7okt9EXcxvoL:b-aak2ru) (no related file):*

Well spotted. I actually overhauled the relationship between property chains and property replay in order to keep the error conditions clean.
</blockquote></details>

The new complexity in the Readme is a sign that we've fallen down a complexity hole.  See below for simplification.
___
*[`src/index.js` line 777 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtjS6ej9qnqoIOf9vVK:-OtkLepX533Yer-WmU2H:breqf0d) ([raw file](https://github.com/meshcat-dev/meshcat/blob/19b509df6c231b5e22598375ab08a248267cf746/src/index.js#L777)):*

As part of the effort to better coordinate with chained properties, this has been clarified, I hope.
</blockquote></details>

I think this can all be *vastly* simpler and more precise -- zero heuristics -- if I've understood correctly.

(1) Don't infer `is_object_placeholder`.  Instead, have passed in via a new constructor argument (defaults to false).  Maybe rename it to "loading_in_progress" or something.

(2) In `set_object_from_json`, don't wait to call `find` until after the mesh has loaded. Call it right away, so that all of the `SceneNode`s are immediately created.  For its final `<object>` node, set `loading_in_progress` to true.

(2b) If the final node already existed (i.e., if we are replacing it), we probably also need to clear out the old geometry immediately (so that properties won't be set on the stale geometry). Depends on exactly how the logic ends up being written.

Now we know _exactly_ which scene nodes are pending, so we can precisely either apply a property, defer it until after the loader finishes, or report an error immediately.  It remains always an error for a `set_property` to refer to a scene node that doesn't exist.

(3) In `SceneNode.set_property`, we should enqueue the property for later application (during set_object) exactly when either

(3a) it's being set on a node where `loading_in_progress` is true, or

(3b) for _recursive_ properties (materials) *only*, on any descendant nodes that are still marked with `loading_in_progress`. It must _not_ be enqueued on their parents, only on the specific in-progress `<object>` leaves.

(4) If you like, one cute representation trick would be to have `pending_properties !== undefined` be the representation of the `loading_in_progress` bool.  In other words, if the `SceneNode` should accept pending properties, then it has a dict to store them.  If not, then the attribute is undefined.  No need to keep two member fields in sync.
___
*[`src/index.js` line 826 at r2](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtkCvlU2F1XOYPjNawL:-OtkCvlU2F1XOYPjNawM:bkrr68h) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f6ac8fc6b54e21dfab4c4a9676fe65afd210eac8/src/index.js#L826)):*
> ```JavaScript
>     }
> 
>     set_property_chain(property, value, target_path, defer_if_missing = false) {
> ```

It doesn't seem to me like `defer_if_missing` is a good name here.  When missing, nothing in _this_ function defers anything. (To me, "defer" means "schedule for later", not "ignore".) In terms of this function, this flag only chooses whether errors are reported vs ignored.
___
*[`src/index.js` line 874 at r2](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtkDLvW9Uq6Q-LVVQnD:-OtkDLvW9Uq6Q-LVVQnE:bsmm1f6) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f6ac8fc6b54e21dfab4c4a9676fe65afd210eac8/src/index.js#L874)):*
> ```JavaScript
>         }
> 
>         if (error_detail != null) {
> ```

nit Wouldn't it be simpler to handle the "ignore errors instead of reporting errors" right here inside this `if`, without any of the other changes in this function?  This is the single point of control for "report and error" so seems to me like a more obvious place to nerf it.

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI made 4 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200)* status: 1 of 5 files reviewed, 2 unresolved discussions (waiting on jwnimmer-tri).

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtjTzjN77Y5k-R_dVjH:-OtkZPBJ5iyNBl_WfKuF:baj0no1) (no related file):*

The new complexity in the Readme is a sign that we've fallen down a complexity hole.  See below for simplification.
</blockquote></details>

Or it could be a measure of previous imprecision that has been made more precise while adding a new feature. I don't see the complexity that you see.
___
*[`src/index.js` line 777 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtjS6ej9qnqoIOf9vVK:-OtkQvPsBV6XIn5WRWB_:b-no0pqw) ([raw file](https://github.com/meshcat-dev/meshcat/blob/19b509df6c231b5e22598375ab08a248267cf746/src/index.js#L777)):*

I think this can all be *vastly* simpler and more precise -- zero heuristics -- if I've understood correctly.

(1) Don't infer `is_object_placeholder`.  Instead, have passed in via a new constructor argument (defaults to false).  Maybe rename it to "loading_in_progress" or something.

(2) In `set_object_from_json`, don't wait to call `find` until after the mesh has loaded. Call it right away, so that all of the `SceneNode`s are immediately created.  For its final `<object>` node, set `loading_in_progress` to true.

(2b) If the final node already existed (i.e., if we are replacing it), we probably also need to clear out the old geometry immediately (so that properties won't be set on the stale geometry). Depends on exactly how the logic ends up being written.

Now we know _exactly_ which scene nodes are pending, so we can precisely either apply a property, defer it until after the loader finishes, or report an error immediately.  It remains always an error for a `set_property` to refer to a scene node that doesn't exist.

(3) In `SceneNode.set_property`, we should enqueue the property for later application (during set_object) exactly when either

(3a) it's being set on a node where `loading_in_progress` is true, or

(3b) for _recursive_ properties (materials) *only*, on any descendant nodes that are still marked with `loading_in_progress`. It must _not_ be enqueued on their parents, only on the specific in-progress `<object>` leaves.

(4) If you like, one cute representation trick would be to have `pending_properties !== undefined` be the representation of the `loading_in_progress` bool.  In other words, if the `SceneNode` should accept pending properties, then it has a dict to store them.  If not, then the attribute is undefined.  No need to keep two member fields in sync.
</blockquote></details>

That's a really dense list. Much of it is statement of solution with which I'm struggling to understand the problems being solved.

As I worked my way through your suggestions (responses, such as  they are, below), one theme kept coming to mind. There seems to be an assumption that one call to set_object creates one `<object>` instance (as child of the named path). That is not generally the case.

Generally, I think this needs more discussion.

-----------------------------------------------------------------

(1) I don't know what you mean by "don't infer". It's not inferred, it's purely the context. When following a node path creates a new child, it automatically sets that to be a place holder. This is the one time that the introduction of a SceneNode can be *known* to be a place holder. Because it was created via traversal and not assignment. The only thing that clears it is actually setting the object (via `SceneObject.set_object()`). That doesn't seem like a heuristic. So, the only difference I can see that your comment implies is that in `create_child()` where we *set* `is_object_placeholder` we would, instead, pass the flag to `add_child` that would in turn pass it to the `SceneNode` constructor.  Is that less complex?

(2a) I'm perhaps being a bit obtuse about `set_object_from_json`. "For its final `<object>` node, set loading_in_progress to true." If the object_json is a glTF file, it won't necessarily be *a* final `<object>` I could be an entire tree. I'll have to think about the implications of what you're saying vis a vis that fact.

(2b) Again, "we know exactly which scene nodes are pending" is not generally true.

(3) Again, handling set_object is not necessarily a single object.
___
*[`src/index.js` line 826 at r2](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtkCvlU2F1XOYPjNawL:-OtkWvr06LMEtQTbnQF7:bb6un1j) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f6ac8fc6b54e21dfab4c4a9676fe65afd210eac8/src/index.js#L826)):*

It doesn't seem to me like `defer_if_missing` is a good name here.  When missing, nothing in _this_ function defers anything. (To me, "defer" means "schedule for later", not "ignore".) In terms of this function, this flag only chooses whether errors are reported vs ignored.
</blockquote></details>

Done
___
*[`src/index.js` line 874 at r2](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtkDLvW9Uq6Q-LVVQnD:-OtkX1PF3OxDDy9M8vqU:b-pbdqz3) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f6ac8fc6b54e21dfab4c4a9676fe65afd210eac8/src/index.js#L874)):*

nit Wouldn't it be simpler to handle the "ignore errors instead of reporting errors" right here inside this `if`, without any of the other changes in this function?  This is the single point of control for "report and error" so seems to me like a more obvious place to nerf it.
</blockquote></details>

Agreed.

jwnimmer-tri (changes requested):
> @jwnimmer-tri made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on SeanCurtis-TRI).

___
*[`src/index.js` line 777 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtjS6ej9qnqoIOf9vVK:-OtkZwc9Al_OP8DAtRqU:b-yuky1q) ([raw file](https://github.com/meshcat-dev/meshcat/blob/19b509df6c231b5e22598375ab08a248267cf746/src/index.js#L777)):*
Yes, I posted a solution instead of trying to spend time itemizing all of the ways in which the current `should_queue_property_for_replay` heuristic fails.  If it helps, we can do a call next week to go over them.

> When following a node path creates a new child, it automatically sets that to be a place holder.

Yes, it's marked as a placeholder if it's new and it's named `<object>`.  However, not all `<objects>` are destined to be replaced by a `set_object` call.  Maybe the user just decided to name something that.  We have no limitation that `<object>` as an element in the path is reserved for use by `set_object`.

The heuristic is that `<object>` is only ever emplaced by `set_object` tacking it onto the end of the path.  That's not an invariant, it's just a guess.

> If the object\_json is a glTF file, it won't necessarily be _a_ final `<object>` I could be an entire tree. I'll have to think about the implications of what you're saying vis a vis that fact.

Good point, I didn't realize that.  In that case my proposal extends to the glTFs root `<object>` being marked as "loading in progress", and then any deferred properties can be applied _recursively_ once it appears.

The key invariant is that `set_object_from_json` must mark _something specific_ as "loading in progress" and that properties must only be deferred for something specifically marked as such.  Trying to guess which properties couldn't be applied probably-because they are still being loaded is too difficult to make accurate.

My original proposal only marked leaves as "loading in progress", but with the "loading can add whole subtrees" we need to amend it to allow marking subtress as well.  But the point is that it must be marked as "this thing is pending because I really do have an async load call targeting it".  Using properties of the path to guess whether something is pending is the main defect.

SeanCurtis-TRI:
*[`src/index.js` line 777 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtjS6ej9qnqoIOf9vVK:-OtkcFqT35696HxPKrcw:bci5lsx) ([raw file](https://github.com/meshcat-dev/meshcat/blob/19b509df6c231b5e22598375ab08a248267cf746/src/index.js#L777)):*

Yes, I posted a solution instead of trying to spend time itemizing all of the ways in which the current `should_queue_property_for_replay` heuristic fails.  If it helps, we can do a call next week to go over them.

> When following a node path creates a new child, it automatically sets that to be a place holder.

Yes, it's marked as a placeholder if it's new and it's named `<object>`.  However, not all `<objects>` are destined to be replaced by a `set_object` call.  Maybe the user just decided to name something that.  We have no limitation that `<object>` as an element in the path is reserved for use by `set_object`.

The heuristic is that `<object>` is only ever emplaced by `set_object` tacking it onto the end of the path.  That's not an invariant, it's just a guess.

> If the object\_json is a glTF file, it won't necessarily be _a_ final `<object>` I could be an entire tree. I'll have to think about the implications of what you're saying vis a vis that fact.

Good point, I didn't realize that.  In that case my proposal extends to the glTFs root `<object>` being marked as "loading in progress", and then any deferred properties can be applied _recursively_ once it appears.

The key invariant is that `set_object_from_json` must mark _something specific_ as "loading in progress" and that properties must only be deferred for something specifically marked as such.  Trying to guess which properties couldn't be applied probably-because they are still being loaded is too difficult to make accurate.

My original proposal only marked leaves as "loading in progress", but with the "loading can add whole subtrees" we need to amend it to allow marking subtress as well.  But the point is that it must be marked as "this thing is pending because I really do have an async load call targeting it".  Using properties of the path to guess whether something is pending is the main defect.
</blockquote></details>

Ah...this time you're referring to `should_queue_property_for_replay` (I think you may have had a typo that led me down the wrong path). That, I agree, is a heuristic. So, all of your proposals are towards making that heuristic disappear in favor of a unambiguous signal.  Let me ruminate on that in conjunction of all the other *stuff*.

SeanCurtis-TRI:
*[`src/index.js` line 777 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtjS6ej9qnqoIOf9vVK:-Otkcjwj0sY_kDZA6bbf:bh1a3sn) ([raw file](https://github.com/meshcat-dev/meshcat/blob/19b509df6c231b5e22598375ab08a248267cf746/src/index.js#L777)):*

Ah...this time you're referring to `should_queue_property_for_replay` (I think you may have had a typo that led me down the wrong path). That, I agree, is a heuristic. So, all of your proposals are towards making that heuristic disappear in favor of a unambiguous signal.  Let me ruminate on that in conjunction of all the other *stuff*.
</blockquote></details>

(BTW I wouldn't happily accepted one case in which the heuristic fails....I didn't need an itemized list.)

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI made 2 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on jwnimmer-tri).

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtjTzjN77Y5k-R_dVjH:-Ou2fzYA8yj7NWKI6Jky:b-5hbxmk) (no related file):*

Or it could be a measure of previous imprecision that has been made more precise while adding a new feature. I don't see the complexity that you see.
</blockquote></details>

I think this is more what you wanted. Eliminating the "you can set the property before you set the object" fluff simplified everything.
___
*[`src/index.js` line 777 at r1](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-OtjS6ej9qnqoIOf9vVK:-Ou2g6TgETfXr4HRoiuk:bb6un1j) ([raw file](https://github.com/meshcat-dev/meshcat/blob/19b509df6c231b5e22598375ab08a248267cf746/src/index.js#L777)):*

(BTW I wouldn't happily accepted one case in which the heuristic fails....I didn't need an itemized list.)
</blockquote></details>

Done

jwnimmer-tri (changes requested):
> @jwnimmer-tri reviewed 3 files and all commit messages, made 6 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200)* status: 3 of 5 files reviewed, 6 unresolved discussions (waiting on SeanCurtis-TRI).

___
*[`Readme.md` line 319 at r4](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou2gsim4-545yj6LC5P:-Ou2gsim4-545yj6LC5Q:bb1al79) ([raw file](https://github.com/meshcat-dev/meshcat/blob/e0b993c36793cd3e617a368e38d0ef54d5641601/Readme.md#L319)):*
nit We need to more clearly underline that "convenience properties" is a term of art that was defined above with a very specific meaning.

We can do that by using a parenthetical instead offsetting with commas.

> Convenience properties (defined above, e.g., `color`) set on ...

But this might be moot; see below.
___
*[`Readme.md` line 317 at r4](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou2ggwHALj9RCwxF5XX:-Ou2ggwHALj9RCwxF5XY:b-areiy3) ([raw file](https://github.com/meshcat-dev/meshcat/blob/e0b993c36793cd3e617a368e38d0ef54d5641601/Readme.md#L317)):*
> ```Markdown
>                         As noted, specifying a `path` that doesn't exist creates that path. However, specifying a property that doesn't exist does *not* create that property.
>                         <p>
>                         If a name in the property chain is missing on an object that has already been created, an error message will be printed to the console and no value will be assigned.
> ```

BTW This paragraph lost the "If the `path` led to the implicit creation of a new folder and object, that pair will still be in place." reminder.  I'm OK if you want to remove that, but wanted to make sure it was on purpose.
___
*[`Readme.md` line 319 at r4](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou2ldUi7ijgdP7969Ee:-Ou2ldUi7ijgdP7969Ef:b-i8o38n) ([raw file](https://github.com/meshcat-dev/meshcat/blob/e0b993c36793cd3e617a368e38d0ef54d5641601/Readme.md#L319)):*
> ```Markdown
>                         If a name in the property chain is missing on an object that has already been created, an error message will be printed to the console and no value will be assigned.
>                         <p>
>                         Convenience properties set on a containing path, such as <code>color</code>, <code>opacity</code>, and <code>modulated_opacity</code>, are immediate subtree operations. They apply to materials that already exist under that path, but they are not remembered for objects added later. Order of operations matters.
> ```

The first few times I read this paragraph, I didn't get it.

Later I realized that *containing path* was the key, trying to draw a distinction versus an `<object>` path.

Properties that target an `<object>` created by `set_object` (which are I guess not a "containing path") are remembered for later, but nothing else is.  We're explaining that in the negative, I suppose because the "queue for set_object pending" is obvious and doesn't need reinforcement.

In any case, this entire exposition is really confusing, even to me who knows how it all works.

One possible idea for adding clarity: bifurcate the explanation of `property` into two wholly-distinct sub-categories: convenience properties, and threejs properties. Each sub-category will have its own explanation of semantics, errors, order of operations, chaining (which is threejs-only, right?) etc. The *only* detail they have in common is "paths are implicitly created", which could be said once at the top (or repeated in both is probably fine, too).
___
*[`Readme.md` line 321 at r4](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou2keVl0UdHDHzpTeU_:-Ou2keVl0UdHDHzpTeUa:bjbpjb4) ([raw file](https://github.com/meshcat-dev/meshcat/blob/e0b993c36793cd3e617a368e38d0ef54d5641601/Readme.md#L321)):*
> ```Markdown
>                         Convenience properties set on a containing path, such as <code>color</code>, <code>opacity</code>, and <code>modulated_opacity</code>, are immediate subtree operations. They apply to materials that already exist under that path, but they are not remembered for objects added later. Order of operations matters.
>                         <p>
>                         More subtly, if the property name chain has an interior name (e.g., the `foo` in `material.foo.color`) that exists but is not an object and does not have properties (such as if `foo` were a `Number`), then, again, an error gets written to the console and no value will be assigned.
> ```

BTW This paragraph is now more separated that is ideal from the _"If a name in the property chain is missing on an object that has already been created, an error message will be printed to the console and no value will be assigned"_ baseline text that it's modifying.
___
*[`src/index.js` line 588 at r4](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou2h8Ro80qChpJgD7mI:-Ou2h8Ro80qChpJgD7mJ:bd2ln5j) ([raw file](https://github.com/meshcat-dev/meshcat/blob/e0b993c36793cd3e617a368e38d0ef54d5641601/src/index.js#L588)):*
> ```JavaScript
>         this.controllers = [];
>         this.on_update = on_update;
>         // Non-null only for auto-created <object> placeholders. The object
> ```

nit Clarity?



_Suggestion:_
```JavaScript
node
```
___
*[`src/index.js` line 609 at r4](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou2hGQt7HIu61wJrniQ:-Ou2hGQt7HIu61wJrniR:b7zfwwh) ([raw file](https://github.com/meshcat-dev/meshcat/blob/e0b993c36793cd3e617a368e38d0ef54d5641601/src/index.js#L609)):*
> ```JavaScript
>         this.object.add(obj);
>         let node = this.add_child(obj);
>         if (name === "<object>") {
> ```

This is still a name-based heuristic. It's a better (more localized) heuristic, but it can still be wrong. If I call `set_property` on `/foo/bar/<object>` without any related `set_object` for `/foo/bar`, then the queued property will stick around forever.

The easy way to be precise is for `set_object_from_json` to either pass a flag that tells us to set this, or for it to directly initialize the dict itself.

We must ensure that queuing is happening exactly when a background loader is running.  Using the name as part of that predicate can never work.

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI made 5 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200)* status: 3 of 5 files reviewed, 5 unresolved discussions (waiting on jwnimmer-tri).

___
*[`Readme.md` line 317 at r4](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou2ggwHALj9RCwxF5XX:-Ou2uLoe1BG2Cyv4yxMB:brx8ckp) ([raw file](https://github.com/meshcat-dev/meshcat/blob/e0b993c36793cd3e617a368e38d0ef54d5641601/Readme.md#L317)):*

BTW This paragraph lost the "If the `path` led to the implicit creation of a new folder and object, that pair will still be in place." reminder.  I'm OK if you want to remove that, but wanted to make sure it was on purpose.
</blockquote></details>

I felt it was largely redundnat w.r.t. the previous sentence, "specifying a `path` that doesn't exist creates that path". It does spell out the proper inference and I'm ok with putting it back (but for now, I was slightly more ok with leaving it out).

However, based on your other comments, I've completely reworked this.
___
*[`Readme.md` line 319 at r4](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou2gsim4-545yj6LC5P:-Ou2um0G8rR3ZSUUtreF:b-4gow27) ([raw file](https://github.com/meshcat-dev/meshcat/blob/e0b993c36793cd3e617a368e38d0ef54d5641601/Readme.md#L319)):*

nit We need to more clearly underline that "convenience properties" is a term of art that was defined above with a very specific meaning.

We can do that by using a parenthetical instead offsetting with commas.

> Convenience properties (defined above, e.g., `color`) set on ...

But this might be moot; see below.
</blockquote></details>

Major document overhaul.
___
*[`Readme.md` line 319 at r4](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou2ldUi7ijgdP7969Ee:-Ou2wJ4Y0gwJuuTxrXnL:b-p0b2yf) ([raw file](https://github.com/meshcat-dev/meshcat/blob/e0b993c36793cd3e617a368e38d0ef54d5641601/Readme.md#L319)):*

The first few times I read this paragraph, I didn't get it.

Later I realized that *containing path* was the key, trying to draw a distinction versus an `<object>` path.

Properties that target an `<object>` created by `set_object` (which are I guess not a "containing path") are remembered for later, but nothing else is.  We're explaining that in the negative, I suppose because the "queue for set_object pending" is obvious and doesn't need reinforcement.

In any case, this entire exposition is really confusing, even to me who knows how it all works.

One possible idea for adding clarity: bifurcate the explanation of `property` into two wholly-distinct sub-categories: convenience properties, and threejs properties. Each sub-category will have its own explanation of semantics, errors, order of operations, chaining (which is threejs-only, right?) etc. The *only* detail they have in common is "paths are implicitly created", which could be said once at the top (or repeated in both is probably fine, too).
</blockquote></details>

Overhaul
___
*[`Readme.md` line 321 at r4](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou2keVl0UdHDHzpTeU_:-Ou3WRI47WOxlsvksLu4:b-p0b2yf) ([raw file](https://github.com/meshcat-dev/meshcat/blob/e0b993c36793cd3e617a368e38d0ef54d5641601/Readme.md#L321)):*

BTW This paragraph is now more separated that is ideal from the _"If a name in the property chain is missing on an object that has already been created, an error message will be printed to the console and no value will be assigned"_ baseline text that it's modifying.
</blockquote></details>

Overhaul
___
*[`src/index.js` line 609 at r4](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou2hGQt7HIu61wJrniQ:-Ou3WjKd3bQnWF7Q-Tsb:bb6un1j) ([raw file](https://github.com/meshcat-dev/meshcat/blob/e0b993c36793cd3e617a368e38d0ef54d5641601/src/index.js#L609)):*

This is still a name-based heuristic. It's a better (more localized) heuristic, but it can still be wrong. If I call `set_property` on `/foo/bar/<object>` without any related `set_object` for `/foo/bar`, then the queued property will stick around forever.

The easy way to be precise is for `set_object_from_json` to either pass a flag that tells us to set this, or for it to directly initialize the dict itself.

We must ensure that queuing is happening exactly when a background loader is running.  Using the name as part of that predicate can never work.
</blockquote></details>

Done

jwnimmer-tri (commented):
This looks more promising now.  I'll dig into more details after lunch.

> @jwnimmer-tri reviewed 1 file and all commit messages, made 2 comments, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-:-Ou3owFf1X9wqsoMYAKP:bvajw27)* status: 1 of 5 files reviewed, 2 unresolved discussions (waiting on SeanCurtis-TRI).

___
*[`Readme.md` line 308 at r5](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou3orwR5Qkoh8RTpSBw:-Ou3orwR5Qkoh8RTpSBx:b-yp97xj) ([raw file](https://github.com/meshcat-dev/meshcat/blob/e97b953f0b60dcc97688382a1795e9e375d8e85a/Readme.md#L308)):*
> ```Markdown
>                         <p>The property name to set, as a string.</p>
> 
>                         <p><strong>Property dialects</strong></p>
> ```

nit Something with the rendering here went sideways:

https://github.com/SeanCurtis-TRI/meshcat/blob/PR_defer_property_replay/Readme.md

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on jwnimmer-tri).

___
*[`Readme.md` line 308 at r5](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou3orwR5Qkoh8RTpSBw:-Ou3xC3u9KS5AA8KOZd2:bh7cq56) ([raw file](https://github.com/meshcat-dev/meshcat/blob/e97b953f0b60dcc97688382a1795e9e375d8e85a/Readme.md#L308)):*

nit Something with the rendering here went sideways:

https://github.com/SeanCurtis-TRI/meshcat/blob/PR_defer_property_replay/Readme.md
</blockquote></details>

Oops. Now better.

jwnimmer-tri (approved):
:lgtm:

I didn't test integration with Drake meshcat, but I assume you've done that.

> @jwnimmer-tri reviewed 4 files and all commit messages, and made 7 comments.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-:-Ou47yOX7b8rOnOKS2sG:b-k11d6m)* status: all files reviewed, 7 unresolved discussions (waiting on SeanCurtis-TRI).

___
*[`Readme.md` line 34 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou43J0C6zpzoqjkicLi:-Ou43J0C6zpzoqjkicLj:b-90u28) ([raw file](https://github.com/meshcat-dev/meshcat/blob/60b48847ec9a3945368ce69b1a3424b6c57fd3d4/Readme.md#L34)):*
> existing children of that container remain

BTW Mostly this was just a re-phrasing of the old text that said effectively the same thing, but possibly it's worth being more specific now?  As I understand it, children that are _container paths_ would remain, but the child (if any) that is a _renderable_ node would be replaced.
___
*[`test/deferred_property_replay.html` line 10 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou47b6CDaLzujShlmya:-Ou47b6CDaLzujShlmyb:b-idly0c) ([raw file](https://github.com/meshcat-dev/meshcat/blob/60b48847ec9a3945368ce69b1a3424b6c57fd3d4/test/deferred_property_replay.html#L10)):*
> Open the developer console (F12).

BTW Is this really necessary?  It didn't seem like it.
___
*[`src/index.js` line 835 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou45NbQ6JCw6kzaWp16:-Ou45NbQ6JCw6kzaWp17:b-vol3wm) ([raw file](https://github.com/meshcat-dev/meshcat/blob/60b48847ec9a3945368ce69b1a3424b6c57fd3d4/src/index.js#L835)):*
> ```JavaScript
>         if (error_detail != null) {
>             if (this.pending_properties !== null) {
>                 // Pending properties exists; this is a place holder. The errors
> ```

nit This says "place holder" (2 words) but other locations say "placeholder" (1 word).  I prefer the latter, but in any case the file should be internally consistent one way or another.
___
*[`Readme.md` line 310 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou44ORj5jcC44sxI1mJ:-Ou44ORj5jcC44sxI1mK:b-ltelzt) ([raw file](https://github.com/meshcat-dev/meshcat/blob/60b48847ec9a3945368ce69b1a3424b6c57fd3d4/Readme.md#L310)):*
> ```Markdown
>                         <ul>
>                             <li><strong>Convenience properties</strong> are short aliases for assigning values to standard scene-graph properties.
>                             <li><strong>Literal properties</strong> are interpreted as literal javascript properties of the object named by the path. The property could be a direct property of the named object (e.g., "type"), or a "chained" property (e.g., "material.map.repeat.x").
> ```

BTW I believe trying to setting literal property on a container node is always an error?  The text could make that more clear.  The current text says `... of the object named ...` but it's not obvious that `object` there means an actually threejs object as opposed to the container node's javascript object.
___
*[`test/deferred_property_replay.html` line 15 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou47er72XKVKHj8fMxX:-Ou47er8-fWVLz1oZpu5:bnx87n6) ([raw file](https://github.com/meshcat-dev/meshcat/blob/60b48847ec9a3945368ce69b1a3424b6c57fd3d4/test/deferred_property_replay.html#L15)):*
> ```HTML
>         </p>
>         <p>
>         You should see four cubes: an opaque green cube, a scaled transparent
> ```

nit It took me a short while to convince myself that what I was seeing on the screen was correct, because the order of the cubes described in the text doesn't match the left-to-right order on the screen.  I think that's working as intended, but it would probably save trouble if the text matched the on-screen layout.
___
*[`test/chained_properties.html` line 14 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou46zd6CWnCbw1FsxZZ:-Ou46zd6CWnCbw1FsxZ_:b-4lus9c) ([raw file](https://github.com/meshcat-dev/meshcat/blob/60b48847ec9a3945368ce69b1a3424b6c57fd3d4/test/chained_properties.html#L14)):*
> ```HTML
>             </p>
>             <p>
>             Open the developer console (F12) for detailed logs if needed.
> ```

BTW When running this test I opened F12 and was trying to see what I could learn there, but then I realized I didn't need to.  Possibly this advice could be amended to something more along the lines of "if you need to debug a FAIL message, open F12 console for more details"?

SeanCurtis-TRI (commented):
Looks clean in Drake.

> @SeanCurtis-TRI made 4 comments and resolved 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-:-Ou4N-ds4WT-lhD0H_wQ:b-padf8s)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on SeanCurtis-TRI).

___
*[`Readme.md` line 34 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou43J0C6zpzoqjkicLi:-Ou4J8pQB4a2MdoSmS3q:bcwtzaq) ([raw file](https://github.com/meshcat-dev/meshcat/blob/60b48847ec9a3945368ce69b1a3424b6c57fd3d4/Readme.md#L34)):*

> existing children of that container remain

BTW Mostly this was just a re-phrasing of the old text that said effectively the same thing, but possibly it's worth being more specific now?  As I understand it, children that are _container paths_ would remain, but the child (if any) that is a _renderable_ node would be replaced.
</blockquote></details>

The word "object" is so overloaded. But, more particularly, I wanted *this* language to match the language above talking about "container" and "renderable" nodes. There's no point in introducing vocabulary if you don't hew to it.
___
*[`Readme.md` line 310 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou44ORj5jcC44sxI1mJ:-Ou4JPjUDvyzbUVZQ2Yj:b-ct8uky) ([raw file](https://github.com/meshcat-dev/meshcat/blob/60b48847ec9a3945368ce69b1a3424b6c57fd3d4/Readme.md#L310)):*

BTW I believe trying to setting literal property on a container node is always an error?  The text could make that more clear.  The current text says `... of the object named ...` but it's not obvious that `object` there means an actually threejs object as opposed to the container node's javascript object.
</blockquote></details>

In fact, it is acceptable to set a literal property on a container node. `visible` is a classic property for toggling on a container node. Admittedly, container nodes don't have very many *useful* properties to set (with the main one hijacked via set_transform or some of the convenience properties). But it's definitely valid. It's one of the reasons that set_property doesn't require or infer `<object>`.
___
*[`src/index.js` line 835 at r6](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou45NbQ6JCw6kzaWp16:-Ou4Js_06UINUrUyRrGh:bb6un1j) ([raw file](https://github.com/meshcat-dev/meshcat/blob/60b48847ec9a3945368ce69b1a3424b6c57fd3d4/src/index.js#L835)):*

nit This says "place holder" (2 words) but other locations say "placeholder" (1 word).  I prefer the latter, but in any case the file should be internally consistent one way or another.
</blockquote></details>

Done

jwnimmer-tri (approved):
+a:@rpoyner-tri for platform review per schedule, please.

> @jwnimmer-tri reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-:-Ou7ZOM62oQz0vjj-Nvw:btx5u2c)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on rpoyner-tri).

rpoyner-tri (approved):
> @rpoyner-tri reviewed 5 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200)* status: all files reviewed, 1 unresolved discussion (waiting on SeanCurtis-TRI).

___
*[`src/index.js` line 870 at r7](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou8-C0q2q5DTkVjpaEj:-Ou8-C0q2q5DTkVjpaEk:bb7phvt) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f691fd9d0829dd9c6b0fb4d89c834cc3a5256675/src/index.js#L870)):*
> ```JavaScript
>         // immediately without waiting for the object to load.
>         let queued_props = this.pending_properties || {};
>         // Clear queue to avoid re-queueing during replay.
> ```

typo



_Suggestion:_
```JavaScript
queuing
```

SeanCurtis-TRI:
I'm currently blocking -- some of the simplifications have broken the texture crawl. I'm going to figure out what's going on before finalizing this. Stay tuned.

rpoyner-tri (approved):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200)* status: all files reviewed, 2 unresolved discussions (waiting on jwnimmer-tri and SeanCurtis-TRI).

jwnimmer-tri (approved):
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200)* status: all files reviewed, 2 unresolved discussions (waiting on rpoyner-tri and SeanCurtis-TRI).

rpoyner-tri (approved):
> @rpoyner-tri made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200)* status: all files reviewed, 2 unresolved discussions (waiting on SeanCurtis-TRI).

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou8L8lDC8vBmBo4yR-0:-Ou8S6SU0yOGJRSRUPO6:b-d0x3dg) (no related file):*

I'm currently blocking -- some of the simplifications have broken the texture crawl. I'm going to figure out what's going on before finalizing this. Stay tuned.
</blockquote></details>

nit Mind setting this discussion to "working", so I can ignore it in my dashboard?

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200)* status: all files reviewed, 2 unresolved discussions (waiting on rpoyner-tri and SeanCurtis-TRI).

___
*[`src/index.js` line 870 at r7](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou8-C0q2q5DTkVjpaEj:-Ou88z326IbUalNfUnvI:b-f4v127) ([raw file](https://github.com/meshcat-dev/meshcat/blob/f691fd9d0829dd9c6b0fb4d89c834cc3a5256675/src/index.js#L870)):*
<details><summary><i>Previously, rpoyner-tri (Rick Poyner (rico)) wrote…</i></summary><blockquote>

typo
</blockquote></details>

Don't let your American background colour your judgement on that spelling.  ;)

rpoyner-tri (approved):
> @rpoyner-tri resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200)* status: all files reviewed, 1 unresolved discussion (waiting on SeanCurtis-TRI).

SeanCurtis-TRI (commented):
> @SeanCurtis-TRI made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on SeanCurtis-TRI).

___
*[a discussion](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-Ou8L8lDC8vBmBo4yR-0:-OuCoTa0ExCBW-SSBc7S:bl41ymw) (no related file):*
<details><summary><i>Previously, rpoyner-tri (Rick Poyner (rico)) wrote…</i></summary><blockquote>

nit Mind setting this discussion to "working", so I can ignore it in my dashboard?
</blockquote></details>

I've worked through the downstream implications. We're all good (and now more fully Americanized in our spelling).

rpoyner-tri (approved):
:lgtm:

> @rpoyner-tri reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200#-:-OuConMx4MTQDCyUCuN2:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on SeanCurtis-TRI).

jwnimmer-tri (approved):
> @jwnimmer-tri reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/meshcat-dev/meshcat/200)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on SeanCurtis-TRI).

